PR #8, opened 2026-06-10
test(hybrid): deterministic regression test for bounded→unbounded marker-loss race

jeffling:
## Summary

Adds an e2e regression test for the checkpoint marker-loss race across the hybrid bounded→unbounded transition. When ClickHouse unsubscribes from `CHECKPOINT_COORDINATOR_CHANNEL` before Kafka subscribes, a Marker broadcast during that gap reaches zero source subscribers, the sink never sees a batch tagged with that epoch, and the coordinator stalls forever. The hybrid-level forwarder (already in `hybrid.rs` on `main`) bridges the gap by keeping a continuous subscription and merging buffered markers into outgoing batches; this test guards against regressing that fix.

## Approach

Two pieces:

### 1. Test-only knob in `advance_to_next_phase`

```
STREAMLING__HYBRID__TRANSITION_DELAY_MS=<ms>
```

When set, `advance_to_next_phase` sleeps for that duration before doing any state work. Production never sets this env var so the path is a no-op outside tests. The natural gap between ClickHouse unsubscribe and Kafka subscribe is ~20–50ms — far too short for a 1s checkpoint interval to land a Marker inside it reliably. Setting the knob to `2500ms` guarantees at least 2 Markers fire inside the empty-subscribers window, making the regression deterministic.

It joins a small family of existing test-only env knobs (e.g. `STREAMLING_E2E_HYBRID_RESUME_CHECKPOINT_INTERVAL_SEC`, `STREAMLING__KAFKA_SOURCE__CONSUMER_FETCH_TIMEOUT_SEC`).

### 2. New e2e test: Scenario 8 in `hybrid_source.rs`

`test_hybrid_source_marker_loss_during_transition` — runs a hybrid pipeline (ClickHouse bounded → Kafka unbounded) with the populated-offset-table production shape, captures the streamling subprocess's stderr, and asserts that at least 2 checkpoint epochs finalize **after** the `Advanced to next phase` log line.

## Reliability

| Code state | Sample size | Post-transition epochs | Outcome |
|---|---|---|---|
| **Broken (forwarder reverted) + knob** | 3 runs | 0, 0, 0 | Deterministic FAIL |
| **Fixed + knob** | 20 runs | 2–3 each | Deterministic PASS (20/20) |

Threshold `≥2` cleanly separates the two states with no overlap. On fixed code, 16/20 runs gave 3 epochs and 4/20 gave 2 — no run dropped to 1 or 0.

## Test plan

- [x] `cargo build -p streamling` — clean
- [x] `just lint` (fmt + clippy `-D warnings`) — clean
- [x] `just e2e-test --test hybrid_source` — full suite passes including scenario 8
- [x] Reliability: 20/20 passes on fixed code
- [x] Negative verification: forwarder reverted → 3/3 deterministic failures with 0 post-transition epochs

## Files

- `crates/streamling-connectors/src/table_providers/hybrid.rs` — adds the test-only env knob inside `advance_to_next_phase`
- `crates/streamling-e2e/tests/hybrid_source.rs` — adds Scenario 8 (`test_hybrid_source_marker_loss_during_transition`)

### DIFF
--- a/crates/streamling-connectors/src/table_providers/hybrid.rs
+++ b/crates/streamling-connectors/src/table_providers/hybrid.rs
@@ -767,6 +767,26 @@ impl HybridTableProvider {
     }
 
     pub async fn advance_to_next_phase(&self) -> DataFusionResult<()> {
+        // Test-only knob: artificially extend the bounded→unbounded
+        // transition window so the marker-loss scenario covered by PR #743
+        // (Scenario 8 in crates/streamling-e2e/tests/hybrid_source.rs) can
+        // be reproduced deterministically instead of relying on the natural
+        // ~20–50ms gap between ClickHouse's unsubscribe and Kafka's
+        // subscribe. Production never sets this env var, so the path is a
+        // no-op outside of tests.
+        if let Ok(ms_str) = std::env::var("STREAMLING__HYBRID__TRANSITION_DELAY_MS")
+            && let Ok(ms) = ms_str.parse::<u64>()
+            && ms > 0
+        {
+            warn!(
+                "Hybrid source '{}': STREAMLING__HYBRID__TRANSITION_DELAY_MS={} \
+                 — sleeping inside advance_to_next_phase. This knob is intended \
+                 for tests only.",
+                self.reference_name, ms
+            );
+            tokio::time::sleep(std::time::Duration::from_millis(ms)).await;
+        }
+
         let mut state = self.state.write().await;
         let from_phase = state.current_phase;
 
--- a/crates/streamling-e2e/tests/hybrid_source.rs
+++ b/crates/streamling-e2e/tests/hybrid_source.rs
@@ -1334,3 +1334,294 @@ sinks:
         count
     );
 }
+
+// ============================================================================
+// Scenario 8: Checkpoint marker loss across bounded→unbounded transition
+// ============================================================================
+
+/// Regression test for the PR #743 scenario: a checkpoint Marker emitted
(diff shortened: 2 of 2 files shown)


rafadaguiar:
I think this can still pass even if the bug is still present.
The test itself has the same race of catching the marker in the middle or not.

crates/streamling-connectors/src/table_providers/hybrid.rs
```diff
@@ -767,6 +767,26 @@ impl HybridTableProvider {
...
 
     pub async fn advance_to_next_phase(&self) -> DataFusionResult<()> {
+        // Test-only knob: artificially extend the bounded→unbounded
+        // transition window so the marker-loss scenario covered by PR #743
+        // (Scenario 8 in crates/streamling-e2e/tests/hybrid_source.rs) can
+        // be reproduced deterministically instead of relying on the natural
+        // ~20–50ms gap between ClickHouse's unsubscribe and Kafka's
+        // subscribe. Production never sets this env var, so the path is a
+        // no-op outside of tests.
+        if let Ok(ms_str) = std::env::var("STREAMLING__HYBRID__TRANSITION_DELAY_MS")
```
rafadaguiar, line 777: a bit weird to have this in the implementation. Maybe move it somewhere else?

jeffling:
My understanding is that the checkpoint interval is set to 1s, where as the transition gap is set to 2.5s (for the test). As a result a checkpoint should be fired during the gap and with the previous but this is trying to catch, it will fail. Do you have suggestions for catching it in a more deterministic fashion?

PR #11, opened 2026-06-12
docs: shutdown investigation — why pipelines hang on SIGTERM and lose tail data

rafadaguiar:
## Summary

Root-cause investigation of the long-standing unclean-shutdown problem: on k8s SIGTERM the pipeline starts shutting down, hangs, and gets SIGKILLed after the grace period — and in Job mode the last buffers are never drained, causing tail data loss. The prior CI-only fix (SourceComplete signals) is also explained: it bolted cancellation onto the checkpoint data plane and drops the final batch before sending it.

This PR adds `shutdown-investigation.md` containing:

- **Part 1 — Plugin pipelines:** three racing SIGTERM handlers, inverted termination order (plugins killed first), blocking bounded-channel sends wedging tokio workers, sink futures cancelled mid-flight by the plugin/sink `select!`, and the never-awaited plugin drain that loses the last ~128 buffered batches on every clean Job completion.
- **Part 2 — Non-plugin pipelines:** the orphaned Kafka lag task whose raw `StreamConsumer` drop triggers the documented `rd_kafka_destroy()` deadlock at runtime teardown (the exact failure `SafeKafkaConsumer` was built to prevent, but the lag consumer isn't wrapped).
- **Part 3 — Immediate fix recommendations** (front-to-back drain, ordered and time-bounded).
- **Part 4 — Proposed architecture:** a `ShutdownController` + `ComponentScope` (cancellation token + tracked spawns) + stop-with-final-checkpoint for state capture + a deadline ladder ending in a hard `process::exit` watchdog, so component authors get two small contracts (`run` reacts to the token, `drain` flushes) and cannot hang the process by construction.
- **Part 5 — Verified assumptions:** incl. that `SourceComplete`/`NUM_RECORDS_BEFORE_STOP` can be replaced by the controller (tests then exercise the *production* drain path on every run), and that prod pods run with the default 30s grace period while the coordinator-stop timeout alone is 30s.
- **Part 6 — Exhaustive hang audit:** 22 distinct hang vectors found across 4 widening sweeps (sweep 4 found nothing new), each with file:line citations and a verdict, clustering into 4 shapes that map 1:1 onto the Part 4 mechanisms. Highlights beyond Parts 1–2: ClickHouse/Postgres sinks retry forever (and the CH HTTP client has no request timeout at all), Kafka sink `flush_producer` loops forever, unbounded startup waits where SIGTERM is unobserved or outright swallowed, un-deadlined WASM calls that no `abort()` can interrupt, and a plugin source stream with no exit condition.

## Why merge this

It's the shared map for the actual fix work: every claim carries a `file:line` reference, the matrix is the checklist for the migration, and §5.4 lists the open product decisions (partial-batch policy, plugin flush budgets, multi-sink limit semantics, admin-API stop endpoint).

## Suggested fix order (follow-up PRs)

1. Wrap the lag consumer / join the lag task (minimal fix for the non-plugin hang).
2. Kafka source: drain-and-send on shutdown instead of `break 'outer`.
3. Await plugin termination with a deadline (fixes Job-mode tail loss).
4. `ShutdownController` + `ComponentScope`; port sources/sinks; delete per-component signal handling, `SourceComplete`, and the `NUM_RECORDS_BEFORE_STOP` transport.
5. Timeouts everywhere the audit says they're missing (CH client, OTLP, retry helpers) + clippy `disallowed_methods` lints + SIGTERM chaos e2e test.

### DIFF
--- a/shutdown-investigation.md
+++ b/shutdown-investigation.md
@@ -1,0 +1,310 @@
+# Investigation: unclean shutdown in production (hang → SIGKILL → data loss)
+
+**Date:** 2026-06-12
+**Symptom:** When k8s sends SIGTERM, the pipeline starts shutting down, then hangs, and k8s SIGKILLs it after the grace period. In Job mode, the last few buffers are not drained, causing data loss at the tail. A prior fix (SourceComplete signals, intended for CI) did not work reliably.
+
+This document covers two distinct failure families: pipelines **with plugins** (source/sink/transform plugins) and pipelines **without plugins** (e.g. Kafka/hybrid → ClickHouse). Both hang, for different root causes.
+
+---
+
+## Part 1 — Pipelines with plugins
+
+### Root cause summary
+
+There is no coordinated shutdown path. Three independent SIGTERM handlers race each other, and the one that kills plugins fires first — killing the *downstream* end of the pipeline while sources and sinks are still pushing data. Sink writes to a dead plugin then block forever on a bounded channel, wedging the tokio runtime. Separately, even on clean Job-mode completion, the process exits without awaiting the plugin sink's drain task, so the last queued buffers are cancelled mid-flush at runtime teardown.
+
+### 1.1 Three competing SIGTERM handlers
+
+| Handler | Location | Behavior on SIGTERM |
+|---|---|---|
+| Plugin shutdown watcher | `crates/streamling-core/src/plugin.rs:110-133` | Immediately calls `terminate_all_plugins()` — enqueues `PluginMsg::Terminate` while data is still flowing |
+| Kafka source handler | `crates/streamling-connectors/src/table_providers/kafka.rs:1288-1298` | `break 'outer` — exits without flushing the partial batch |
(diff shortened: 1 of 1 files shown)


jeffling:
Does this investigation cover the case of a job pipeline with multiple sources, one of which terminates before the others?

rafadaguiar:
> Does this investigation cover the case of a job pipeline with multiple sources, one of which terminates before the others?

It should work regardless of that, but I did some digging because what you mentioned and I did find a bug in checkpointing. I believe this fell through the cracks mostly because these bounded cases were introduced after the checkpointing was already done.

If a bounded source ends early it stops participating in the checkpointing and the coordinator keeps waiting for it (epochs don't make progress). We need to unregister the bounded source from the checkpoint coordinator once it's done regardless of the app receiving a shutdown request. I can do that in parallel since I'll be in the neighbourhood of that code anyways.

jeffling:
[goldsky-io/streamling-private#18](<https://github.com/goldsky-io/streamling/pull/18>)<br><br><br>@rafadaguiar here is my investigation into it

PR #16, opened 2026-06-16
feat(clickhouse): adaptive sort-key-range source pagination

jeffling:
## Why

The ClickHouse source query ordered its scan by the sorting key, which forces read-in-order on the main table and makes ClickHouse skip any matching `PROJECTION` (read-in-order is unsupported on projections). On a 1M-row table with an `address` projection, the `ORDER BY` query read **62 marks** and used no projection; the same query without `ORDER BY` used the projection and read **2 marks**.

We had the `ORDER BY` to ensure consistent pagination, and to allow for pagination inside of each sorting key range (block_number). ie block_number 5 might have 10k items, which would have overflowed.

However, our default page size is 10m. it's unlikely that our top-level sorting key has that many items. In case where memory is an issue, we should be redoing the sorting key to be more granular.

## What

- **Drop the `ORDER BY`** from the source query. Pagination is now driven by disjoint half-open ranges on the first sorting key, so determinism comes from the predicate, not row order.
- **`RangeController`** (new) owns the range cursor and adapts the range width:
  - `LIMIT page_size + 1` tripwire — an overflowing range is discarded and re-read smaller, with no duplicates (the response is buffered before emit).
  - up-front `count()` probe sizes the first range from observed density.
  - soft time budget shrinks the range before the hard query timeout fires.
  - only a completed page advances the cursor; overflow and timeout re-read the same range.
- **Checkpoint cursor** is the range start — restart re-reads at most the in-flight range (at-least-once, covered by existing sink dedup).
- **Breaking:** config `block_range` renamed to `sort_key_range` (`STREAMLING__CLICKHOUSE_SOURCE__SORT_KEY_RANGE`).

## Testing

- **Unit** (`range_controller`): emitted ranges tile the key space with no gaps and every row exactly once, under overflow, injected timeouts, and width down/up rescaling. 232 connectors unit tests pass.
- **e2e** (`clickhouse_source`, real ClickHouse): boundary, keyset, multi-range overflow, and checkpoint-across-sparse-ranges all pass (4/4).
- Verified on a real table + projection that the no-`ORDER BY` query selects the projection (2 marks vs 62 with `ORDER BY`).

---

## Version-aware source deduplication (added)

The no-`ORDER BY` scan emits raw physical rows, which for `ReplacingMergeTree` tables means duplicate versions of each `ORDER BY` key reach downstream in scan order — and the existing position-based dedup (last row wins) picks the wrong winner. This change dedupes **at the source**, by max version, before emission.

- **Infer the version column** from `system.tables.engine_full` (`parse_replacing_merge_tree_dedup`). Tokenizing is quote-aware, so commas inside the `'/clickhouse/...'` path literal don't split; the quoted path/replica args are dropped, leaving the bare `version` and `is_deleted` identifiers. Handles `Shared` / `Replicated` / plain `ReplacingMergeTree`; returns nothing for non-Replacing engines.
- **Per-page dedup by max version** (`deduplicate_record_batches_by_version`). Each fully-read page is coalesced and collapsed by `max(insert_timestamp)` keyed on the table's full `ORDER BY`. All versions of a key share the first sorting key, so they always land in one page — per-page coalescing is complete and correct (no cross-batch state needed). Ties break by position (ReplacingMergeTree merge order).
- **Tombstones (`FINAL` semantics):** keys whose max-version row is a delete (`_gs_op='d'`) are dropped entirely. To propagate deletes downstream instead, omit the tombstone rule (one line).
- Reuses the existing hash + real-PK-equality collision handling (STRM-5990). Tombstone match is type-safe via a literal array + `make_comparator` (no per-row `ScalarValue` allocation).

### Force-include the version column (activation fix)

Initially dedup was gated on the inferred version column being present in the fetched schema. That silently disabled it for the **hybrid source** path: `ClickHouseSchemaAdapter::get_columns` projects ClickHouse to the unbounded source's (Kafka) target schema, which excludes ClickHouse housekeeping columns like `insert_timestamp` and `is_deleted` — so the inferred version column wasn't selectable, and the bounded backfill emitted raw, undeduped rows. Same trap for any direct source with explicit `columns`.

Fix: decouple the **internal scan schema** from the **external provider schema**.

- `new_source` appends the inferred version column to the scan column list when the configured columns omit it (`*` already selects every column, so nothing is added). On the rare invalid-version fetch_schema failure we fall back to the configured columns with dedup disabled.
- `SourceParams.project_out_version_index: Option<usize>` records the column to strip — set only when we force-included it.
- `execute()` projects that column out of every emitted batch via `project_out_column`, so the stream's external schema (and the hybrid union's schema contract) is unchanged.

The inference log now reports whether force-inclusion was needed, and the disabled-case log distinguishes "version column not selectable" (invalid inference) from "no Replacing engine."

### Testing (dedup)

- **Unit** — 6 version-aware dedup tests (max-version-wins-not-position, tie→later, tombstone-winner-drops-key, tombstone-non-winner-kept, missing-version-errors, cross-batch): `streamling-common` `utils::dedup` — 30 pass.
- **Unit** — 6 engine-parser tests: `streamling-connectors` lib — 241 pass.
- **e2e regression** — `test_clickhouse_source_replacing_dedup_when_version_column_not_selected` (`clickhouse_source.rs`): creates a `ReplacingMergeTree(insert_timestamp, is_deleted)` table, inserts 9 raw rows across 5 keys covering five dedup scenarios (single version, latest-wins, **position-vs-version trap**, tombstone-winner-dropped, tombstone-non-winner-kept), runs the pipeline with **`columns: "block_number,id,payload"` deliberately omitting the version column**, and asserts: row count = 4 (FINAL); `'c_new'` wins the position-vs-version case; key `'d'` is absent; key `'e'` survives as `'e_alive'`; postgres table has neither `insert_timestamp` nor `is_deleted`. Compiles + lints clean here; must be run locally via `just e2e-test test_clickhouse_source_replacing_dedup_when_version_column_not_selected` (the in-CI cluster's pods are unhealthy in this worktree).
- `just fix && just lint` clean.

### Note

Coalescing the full page adds one in-memory copy of the page (the raw page is already buffered pre-emit, so peak is ~2× page data). Tune `page_size` down on duplicate-dense tables if memory-bound; correctness is unaffected — `RangeController` still rescans on overflow.

### DIFF
--- a/crates/streamling-common/src/utils/dedup.rs
+++ b/crates/streamling-common/src/utils/dedup.rs
@@ -342,6 +342,7 @@ pub fn deduplicate_record_batches_by_version(
 
     deduplicate_record_batch_by_version(&concatenated, primary_key, version_column, tombstone)
 }
+
 /// Feed an array value at `index` into `hasher`. Used to bucket rows by their
 /// primary-key hash; equality is verified separately via `make_comparator` so
 /// `u64` hash collisions do not drop rows (STRM-5990).
--- a/crates/streamling-common/src/utils/dedup.rs
+++ b/crates/streamling-common/src/utils/dedup.rs
@@ -265,12 +265,23 @@ pub fn deduplicate_record_batch_by_version(
         let bucket = buckets.entry(hash).or_default();
         let mut updated = false;
         for entry in bucket.iter_mut() {
-            let (existing, _) = *entry;
+            let (existing, existing_is_tomb) = *entry;
             if pk_eq(existing, row_index) {
-                // Replace when the new row's version >= the current winner's.
-                // `>=` (existing is NOT greater) makes ties resolve to the later
-                // position, matching ReplacingMergeTree merges.
-                if !version_cmp(existing, row_index).is_gt() {
+                // Keep the higher version. On a tie, prefer the tombstone: a
+                // delete at the same version supersedes the live row, so the
+                // key is dropped — ReplacingMergeTree FINAL with an `is_deleted`
+                // flag. The scan has no ORDER BY, so without this the winner is
+                // order-dependent and a live row scanned after the delete
+                // survives, silently losing the deletion. When neither side is a
+                // tombstone (e.g. no tombstone rule), ties keep the later row,
+                // matching merge order.
+                let should_replace = match version_cmp(existing, row_index) {
+                    o if o.is_lt() => true,  // new row has the higher version
+                    o if o.is_gt() => false, // current winner has the higher version
+                    // tie: tombstone wins; if both/neither, later position wins
+                    _ => !existing_is_tomb || row_is_tomb,
+                };
+                if should_replace {
                     *entry = (row_index, row_is_tomb);
                 }
                 updated = true;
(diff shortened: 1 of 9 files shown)


jeffling:
Need to run this across some prod datasets first before merging.

rafadaguiar (approved):

rafadaguiar (approved):

jeffling:
There is a real overflow issue if the columns overflow. due to the large page size (since we don't have a row limit anymore) it's possible. 

Adding detection so the page size reducing can handle that scenario before it reaches a downstream dedup or unnest (which triggers it due to a datafusion issue with a hardcoded int32 buffer size somewhere inthe code path)

rafadaguiar (approved):

PR #1, opened 2026-06-08
docs: reposition README as streaming runtime with onboarding guide

jeffling:
## Summary

- Framework -> Runtime for clarity
- Add onboarding sections: when to use, how it works, quick start (with streamling.dev install script), and common patterns including plugin examples
- Document hybrid and ClickHouse bounded sources for batch/job-mode pipelines
- Replace examples generic CRM/orders/events patterns throughout sample YAML

## Test plan

- [x] README renders correctly on GitHub
- [x] Internal anchor links resolve (Quick start, Plugin System, connector sections)

PR #21, opened 2026-06-17
fix: u256/i256 correctness in CASE expressions and ClickHouse hybrid

jeffling:
## Summary

Three related correctness fixes for 256-bit integer handling, each independently reproducible and unit-tested.

### 1. CASE expressions over u256/i256 dropped type/metadata

**Symptom.** A `CASE` whose branches mix a bigint column with a literal (e.g. `ELSE 0`) failed DataFusion type-coercion (`FixedSizeBinary(32)` vs `Int64`) and terminated the stream. A `CASE` whose branches were all bigint planned, but the output field lost the `streamling.u256`/`i256` extension metadata, so the Postgres sink fell back to `BYTEA` instead of `numeric(78,0)`.

**Fix.** The SQL preprocessor now recurses into `CASE`: arithmetic branches get rewritten in place, literal branches are wrapped in `to_<kind>()` to share the `FixedSizeBinary(32)` type, and the whole `CASE` is wrapped in `to_<kind>()` to re-declare the extension metadata. `to_u256` grows a `FixedSizeBinary(32)` passthrough so the outer wrap is a no-op for already-encoded values.

Bigint overflow errors now include the operand values so an offending row can be identified without guessing.

### 2. ClickHouse hybrid read path stored u256/i256 little-endian as big-endian

**Symptom.** Multiplying u256 columns sourced from a hybrid ClickHouse provider raised \"U256 multiplication overflow\" on values that obviously fit. `u256_to_string` reported astronomic values for known-small wei amounts.

**Cause.** The sink already reversed bytes (`normalize_batch_for_clickhouse`, BE→LE) because ClickHouse stores `UInt256`/`Int256` little-endian. The read path never reversed back. ClickHouse emits `UInt256` as Arrow `FixedSizeBinary(32)` in little-endian; streamling reads big-endian. A logical \`1\` arrived as \`2^248\`, and any multiplication of wei-scale values overflowed \`U256::MAX\`.

**Fix.** New `ClickHouseClient::normalize_batch_from_clickhouse` mirrors the sink-side normalizer: for each column whose target field has u256/i256 metadata but whose incoming field lacks it, reverse the 32 bytes and adopt the target field. Applied in `HybridSourceExec::execute`'s batch-receiving loop. No-op on Kafka/Avro batches (already metadata-tagged + BE).

### 3. Normalizer broke List columns whose inner field names differ

**Symptom (prod).** After fix #2 shipped, `obib-case4-baseline` failed at column 28 with:
> `Invalid argument error: column types must match schema types, expected List(Field name: \"element\", ...) but found List(Field name: \"item\", ...)`

**Cause.** The first cut of `normalize_batch_from_clickhouse` rebuilt every column against \`target_schema\` verbatim. The hybrid layer's \`is_compatible_data_type\` deliberately tolerates ClickHouse's \`item\` vs Kafka/Avro's \`element\` for List inner fields — but \`RecordBatch::try_new\` doesn't.

**Fix.** Only swap the \`Field\` for columns whose bytes we actually transform (u256/i256 byte reversal). All other columns keep their incoming \`Field\`. The output schema mixes incoming and target fields per-column, matching the hybrid's relaxed compatibility rules. A unit regression reproduces the exact prod error before the fix and asserts the same batch flows through afterwards.

## Test plan

- [x] \`cargo nextest run -p streamling-common -p streamling-core -p streamling-connectors\` — all new and existing unit tests pass
- [x] \`just lint\` clean
- [x] New unit tests pin behavior:
  - \`u256\`/\`u256_ops\`: 6 endianness pin tests (BE storage, mis-endian explodes, wei-squared safe, MAX boundary)
  - \`bigint_sql_preprocessor\`: 6 CASE rewrite tests (literal branch, u256/u256, arithmetic branch, non-u256 left alone, i256 parallel)
  - \`clickhouse\`: 4 normalize_batch tests (LE reversal, metadata-tagged passthrough, no-op without bigint, **List inner-name preservation**)
  - \`session\`: 2 end-to-end CASE-over-u256 tests (DataFusion coercion + metadata preservation)
- [x] New e2e test \`test_hybrid_clickhouse_uint128_widened_to_u256_preserves_value\` exercises the full hybrid path (CI runs against k3s)
- [x] Verify against \`obib-case4-baseline\` on the new image (\`fix-case-bigint-coercion-latest\` after this lands)

### DIFF
--- a/crates/streamling-common/src/functions/u256_ops.rs
+++ b/crates/streamling-common/src/functions/u256_ops.rs
@@ -42,6 +42,7 @@ impl ToU256Func {
                     TypeSignature::Exact(vec![DataType::UInt16]),
                     TypeSignature::Exact(vec![DataType::Int8]),
                     TypeSignature::Exact(vec![DataType::UInt8]),
+                    TypeSignature::Exact(vec![DataType::FixedSizeBinary(32)]),
                 ],
                 Volatility::Immutable,
             ),
@@ -201,6 +202,10 @@ impl ScalarUDFImpl for ToU256Func {
                     let bytes = u256::u256_to_bytes(&value);
                     builder.push(Some(bytes.to_vec()));
                 }
+            }
+            DataType::FixedSizeBinary(32) => {
+                // Passthrough: input is already U256
+                return Ok(ColumnarValue::Array(array));
             }
             _ => {
                 streamling_user_bail!(
@@ -482,6 +487,46 @@ mod tests {
         } else {
             panic!("Expected array result");
         }
+    }
+
+    #[test]
+    fn test_to_u256_fixed_size_binary_passthrough() {
+        // to_u256 over already-encoded u256 bytes (FixedSizeBinary(32)) must pass
+        // the value through unchanged. The preprocessor relies on this to wrap
+        // CASE results so the u256 extension metadata is re-declared on the output
+        // field (mirrors to_i256, which already supports this).
+        let func = ToU256Func::new();
+        let bytes = u256::u256_to_bytes(&U256::from(12345u64));
+        let input = FixedSizeBinaryArray::try_from_sparse_iter_with_size(
+            vec![Some(bytes.to_vec()), None].into_iter(),
+            32,
+        )
+        .unwrap();
+
+        let args = ScalarFunctionArgs {
+            args: vec![ColumnarValue::Array(Arc::new(input))],
+            arg_fields: vec![Arc::new(arrow_schema::Field::new(
+                "value",
+                U256Type::new(),
+                true,
+            ))],
+            number_rows: 2,
+            return_field: Arc::new(arrow_schema::Field::new("result", U256Type::new(), true)),
+        };
+
+        let result = func.invoke_with_args(args).unwrap();
+        let ColumnarValue::Array(arr) = result else {
+            panic!("expected array result");
+        };
+        let binary = arr
+            .as_any()
+            .downcast_ref::<FixedSizeBinaryArray>()
+            .expect("result should be FixedSizeBinary");
+        assert_eq!(binary.len(), 2);
+        let mut recovered = [0u8; 32];
+        recovered.copy_from_slice(binary.value(0));
+        assert_eq!(u256::bytes_to_u256(&recovered), U256::from(12345u64));
+        assert!(binary.is_null(1), "null should pass through");
     }
 
     #[test]
--- a/crates/streamling-common/src/functions/u256_ops.rs
+++ b/crates/streamling-common/src/functions/u256_ops.rs
@@ -530,6 +530,72 @@ mod tests {
     }
 
     #[test]
+    fn test_u256_to_string_decodes_bytes_as_big_endian() {
+        // u256_to_string is the only public path from the FixedSizeBinary(32)
+        // buffer to a human-readable decimal — if it ever switched to little-
+        // endian decoding (or a producer ever wrote LE bytes), value 1 would
+        // surface as 2^248 and most multiplications would overflow.
+        let bytes = u256::u256_to_bytes(&U256::from(1u64));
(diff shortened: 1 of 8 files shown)


kashive:
AI review found an issue; seems like worth addressing:


**The risk:** `normalize_batch_from_clickhouse` has two positional assumptions that, if violated, cause it to silently emit **wrong data** rather than error:

```rust
// guard
if batch.num_columns() != target_schema.fields().len() {
    return Ok(batch.clone());   // <- skips reversal entirely
}
// pairing
batch.columns().iter()
    .zip(original_schema.fields())
    .zip(target_schema.fields())   // <- pairs by position, not by name
```

- If the incoming batch is **narrower** than `target_schema` (`self.provider.schema`, the full schema), the guard trips and u256/i256 columns pass through **little-endian, unreversed** → a logical `1` becomes `2^248` → downstream multiply overflows. This is exactly the bug the PR fixes, re-introduced for that batch shape.
- If columns are **reordered** but the count coincidentally matches, the positional `zip` reverses/validates each column against the wrong target field.

**Why it's NOT currently reachable (the mitigating finding):** The bounded ClickHouse source ignores pushed-down projection. Its `scan` takes `_projection` (underscore-prefixed, unused) and builds `ClickHouseSourceExec` from `self.schema.clone()`:

```rust
async fn scan(
    &self,
    _state: &dyn Session,
    _projection: Option<&Vec<usize>>,
    ...
) -> Result<Arc<dyn ExecutionPlan>> {
    ...
    let clickhouse_source_exec = Arc::new(ClickHouseSourceExec {
        cached_properties: PlanProperties::new(
            EquivalenceProperties::new(self.schema.clone()),
            ...
```

So the bounded phase always emits **full-width, full-order** batches matching `self.provider.schema`. The guard passes, the positional pairing is correct, and reversal happens. Net: **today this is a latent fragility / defense-in-depth gap, not an active data-corruption bug.**

**Why it's still worth addressing:** The correctness of a data-integrity transform rests on an unstated, non-local invariant ("the bounded source never projects or reorders"). That invariant lives in a *different* file (`clickhouse.rs` ignoring projection) and could be changed by someone who has no idea this normalizer depends on it — e.g. adding real projection pushdown to the ClickHouse source for performance, or adding a second bounded source type that *does* honor projection. The failure mode would be silent (wrong numbers in a sink), not a crash.

## Suggested actions

In rough priority order:

1. **Match columns by name, not position.** Build a lookup of `target_schema` fields by name and, for each incoming column, find its target field by name. This removes the reorder hazard entirely and makes a subset projection safe (each present column still gets the right target field). Low effort, removes two assumptions at once.

2. **Make a genuine count/shape mismatch loud instead of silent.** Rather than `return Ok(batch.clone())` on count mismatch, either (a) handle it correctly via name-matching per #1, or (b) if a target field with u256/i256 metadata has no corresponding incoming column, return a `StreamlingError` with context. A wrong-shape batch reaching this function is a programming error; failing fast beats emitting corrupted u256 values.

3. **Add a regression test for the subset/reordered case.** A unit test that feeds a 2-of-3-column batch (one of them a LE u256 column) and asserts the u256 column is still reversed and tagged. This pins the behavior so a future projection-pushdown change to the ClickHouse source can't silently regress it. This mirrors how fix #3's List test pinned the prod failure.

4. **Document the cross-file invariant.** A short comment on the guard noting that it currently relies on the bounded source emitting full-width batches, and that honoring projection upstream would require this function to match by name. Cheap insurance for the next reader.

5. **(Optional, broader) Reconcile `HybridSourceExec::schema()` with projection.** `schema()` returns the full provider schema while `execute` forwards a projection to the inner scan. Today that's consistent only because ClickHouse ignores projection. If you ever want real projection pushdown through the hybrid, `schema()` would need to return the projected schema and the normalizer would need #1. Worth a tracking note even if out of scope for this PR.

kashive (approved):
Overall, LGTM 🚀 

AI did find one issue that maybe worth looking into.

PR #22, opened 2026-06-17
Upgrade DataFusion 49 → 54 (arrow 55 → 58.3)

rafadaguiar:
## Summary

Upgrades the DataFusion ecosystem **49.0.2 → 54.0.0** and **arrow 55.2.0 → 58.3.0** across the workspace.

Verified: whole workspace compiles (lib + tests + bin), **856 lib unit tests pass**, and the streamling-e2e suite is green.

## Mechanical migration (per the df50–54 upgrade guides)

- **UDF traits** now require `PartialEq + Eq + Hash` (`DynEq`/`DynHash`) and dropped `as_any`: derive the traits, or hand-impl identity for UDFs holding non-comparable state (`DynamicTableCheckFunc`, `PluginScalarUdf`, the EVM/IDL decoders); remove `as_any` overrides; callers use `downcast_ref` directly.
- **`ScalarFunctionArgs`** gained `config_options`.
- **`ExecutionPlan`**: `properties()` → `&Arc<PlanProperties>`; `statistics()` → `partition_statistics(_, ) -> Result<Arc<Statistics>>`.
- **arrow 55→58**: `Decimal32`/`Decimal64` match arms; `ColumnStatistics::byte_size`; `UnionFields::try_new`; sqlparser `Cast.array`; new `MetricValue` variants; `DFSchema → Schema` via `as_arrow()`; `FilterExec::projection()` now returns `Option<Arc<[usize]>>`.

## Avro

df54 rewrote `datafusion-datasource-avro` on top of `arrow-avro` and dropped the `apache-avro`-based `to_arrow_schema`. streamling's Avro stack is built on `apache-avro`, so the converter (~260 LOC) was **vendored** into `streamling-common`, and the `datafusion-datasource-avro` dependency was removed.

## Scan-sharing + `UNION ALL` (real df54 regression — fixed)

Pipelines reading the same source twice (`… FROM s UNION ALL SELECT … FROM s`) enable scan sharing and regressed on df54. Two issues:

1. **Planning** — df54 strictly asserts an extension node's physical schema matches its logical schema. The shared `BroadcastingExec` emitted the full source schema while the logical `Checkpointable` node was projected. Fix: `BroadcastingExec` is now projection-aware (shared source broadcasts full rows; each consumer projects via metadata-preserving `RecordBatch::project`).
2. **Runtime hang** — the `Checkpointable` planner inserted a **stock `RepartitionExec`** to coalesce the multi-partition `UNION ALL` input. Stock partition operators rebuild batches with a metadata-less schema, dropping the checkpoint marker streamling carries in batch schema metadata → the sink never acked → hang. Fix: `CheckpointableExec` now coalesces its input partitions itself (marker-preserving) and declares `UnspecifiedDistribution`; the stock `RepartitionExec` is removed.

---

> [!NOTE]
> **High Risk**
> Major DataFusion/Arrow upgrade touches query planning, physical operators, Avro decoding, and streaming checkpoint semantics; regressions are most likely in SQL edge cases, scan-shared pipelines, and sink schema registration.
> 
> **Overview**
> This PR bumps the workspace query stack to **DataFusion 54** and **Arrow 58.3** (`object_store` **0.13**, `datafusion-ffi` **54**), with a large `Cargo.lock` refresh across the Arrow/DataFusion dependency tree.
> 
> **DataFusion 54 API migration** is applied across custom UDFs and operators: `ScalarUDFImpl`/`AggregateUDFImpl` types get **`PartialEq`/`Eq`/`Hash`**, **`as_any` overrides are removed**, `ScalarFunctionArgs` literals add **`config_options`**, and `ExecutionPlan` now uses **`Arc<PlanProperties>`**, **`partition_statistics`**, and direct **`downcast_ref`**. Arrow **58** additions (e.g. **`Decimal32`/`Decimal64`**) are handled in Avro writer paths and vendored schema conversion.
> 
> **Avro** no longer depends on `datafusion-datasource-avro`: **`to_arrow_schema` and helpers are vendored** into `streamling-common`, with **`Schema::Ref`** failing cleanly, **local-timestamp** logical types mapped in schema + value resolution, and a decode regression test.
> 
> **UDF return-type strictness (df54):** `array_filter` (and related list UDFs) now **declare non-nullable `"item"` list elements** to match what execution already produced—fixes plan-time assertion failures without changing batch bytes.
> 
> **Scan-sharing + `UNION ALL` over the same source** is fixed for df54: **projection-aware `BroadcastingExec`** and **`CheckpointableExec` coalescing** that preserves checkpoint batch metadata (avoids schema mismatch at plan time and hangs from stock `RepartitionExec` dropping markers).
> 
> `DATAFUSION_54_UPGRADE_NOTES.md` documents the migration, e2e outcomes, and follow-ups (plugins lockstep, optional arrow-avro path).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 656b8a435ac026d6b55a0c292f2df1846c072128. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/streamling-common/src/functions/array_filter.rs
+++ b/crates/streamling-common/src/functions/array_filter.rs
@@ -15,7 +15,6 @@ use datafusion::logical_expr::{
     ColumnarValue, ReturnFieldArgs, ScalarFunctionArgs, ScalarUDFImpl, Signature, TypeSignature,
     Volatility,
 };
-use std::any::Any;
 use std::sync::Arc;
 
 /// array_filter(list<struct>, field_name_utf8, value_utf8) -> list<struct>
@@ -23,7 +22,7 @@ use std::sync::Arc;
 /// Filters each list, keeping only elements where the named struct field (Utf8)
 /// equals the provided value (Utf8), or both are NULL. If the list is null, the result is null.
 /// If the field is missing or not Utf8, an error is returned.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq, Hash)]
 pub struct ArrayFilterFunc {
     signature: Signature,
 }
@@ -747,10 +746,6 @@ pub fn eval_array_filter(
 }
 
 impl ScalarUDFImpl for ArrayFilterFunc {
-    fn as_any(&self) -> &dyn Any {
-        self
-    }
-
     fn name(&self) -> &str {
         "array_filter"
     }
@@ -771,7 +766,15 @@ impl ScalarUDFImpl for ArrayFilterFunc {
         }
         let input_field = &args.arg_fields[0];
         let list_dt = match input_field.data_type() {
-            DataType::List(field) => DataType::List(field.clone()),
+            // The execution path rebuilds the list element field as a non-nullable
+            // "item" field (see `out_field` in the filter implementation). df54 strictly
+            // asserts the produced array type matches the type promised here, so mirror
+            // that element field exactly rather than cloning the input field.
+            DataType::List(field) => DataType::List(Arc::new(SchemaField::new(
+                "item",
+                field.data_type().clone(),
+                false,
+            ))),
             other => {
                 streamling_user_bail!(
                     "array_filter expects first argument to be a list/array, got {:?}",
@@ -855,6 +858,7 @@ mod tests {
             arg_fields: vec![arg0_field.into(), arg1_field.into(), arg2_field.into()],
             number_rows: 2,
             return_field: return_field.into(),
+            config_options: ::std::sync::Arc::new(::datafusion::config::ConfigOptions::default()),
         };
 
         let result = func.invoke_with_args(args).unwrap();
@@ -943,6 +947,7 @@ mod tests {
             arg_fields: vec![arg0_field.into(), arg1_field.into(), arg2_field.into()],
             number_rows: 2,
             return_field: return_field.into(),
+            config_options: ::std::sync::Arc::new(::datafusion::config::ConfigOptions::default()),
         };
 
         let result = func.invoke_with_args(args).unwrap();
@@ -1005,6 +1010,7 @@ mod tests {
             arg_fields: vec![arg0_field.into(), arg1_field.into(), arg2_field.into()],
             number_rows: 2,
             return_field: return_field.into(),
+            config_options: ::std::sync::Arc::new(::datafusion::config::ConfigOptions::default()),
         };
 
         let result = func.invoke_with_args(args).unwrap();
--- a/crates/streamling-common/src/functions/zip_arrays.rs
+++ b/crates/streamling-common/src/functions/zip_arrays.rs
@@ -6,7 +6,6 @@ use datafusion::common::Result;
 use datafusion::logical_expr::{
     ColumnarValue, ReturnFieldArgs, ScalarFunctionArgs, ScalarUDFImpl, Signature, Volatility,
 };
-use std::any::Any;
 use std::sync::Arc;
 
 use crate::functions::util::{get_all_args_as_type, validate_arg_count};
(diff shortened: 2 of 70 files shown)


rafadaguiar:
Tested in prod with a UNION ALL pipeline.

jeffling:
i'd be interested in understanding if this changes kafka throughput, the new avro reader is supposed to be more efficient. we might have to explicitly use it though.

rafadaguiar:
We currently don't use it. But def should look into that.
I'll create an issue.
https://github.com/goldsky-io/streamling/issues/25

crates/streamling-common/src/functions/zip_arrays.rs
```diff
@@ -74,10 +69,13 @@ impl ScalarUDFImpl for ZipArraysFunc {
...
         for (i, arg_type) in arg_types.iter().enumerate() {
             if let DataType::List(field) = arg_type {
+                // The execution path builds these struct fields as nullable (zipped
+                // elements may be absent); df54 asserts produced == promised, so the
+                // declared nullability must match the produced one (always `true`).
                 field_types.push(Field::new(
                     format!("f{}", i),
                     field.data_type().clone(),
-                    field.is_nullable(),
+                    true,
```
xiaom, line 78: In this case, do we produce a non-nullable column for df49? And will it change the sink schema?

crates/streamling-common/src/functions/array_filter.rs
```diff
@@ -771,7 +766,15 @@ impl ScalarUDFImpl for ArrayFilterFunc {
...
         let list_dt = match input_field.data_type() {
-            DataType::List(field) => DataType::List(field.clone()),
+            // The execution path rebuilds the list element field as a non-nullable
+            // "item" field (see `out_field` in the filter implementation). df54 strictly
+            // asserts the produced array type matches the type promised here, so mirror
+            // that element field exactly rather than cloning the input field.
+            DataType::List(field) => DataType::List(Arc::new(SchemaField::new(
+                "item",
+                field.data_type().clone(),
+                false,
```
xiaom, line 776: similar to https://github.com/goldsky-io/streamling/pull/22#discussion_r3495199321, will the non-nullable item field introduce any sink schema change?

crates/streamling-common/src/functions/zip_arrays.rs
rafadaguiar: (This covers the `array_filter` thread too — same mechanism.)

Checked every sink (core + plugins). This is a **declaration-only** change: df49's execution path already produced a non-nullable element named `"item"` (`out_field` in the impls, unchanged by this PR) — df54 just makes the *declared* return type match what was always produced, since df54 asserts produced == promised and df49 silently tolerated the mismatch. So the runtime `RecordBatch` — and the data actually written — is identical to df49.

Sink-schema impact for the `zip_arrays` / `array_filter` `List<Struct>` columns:

- **Kafka (Avro): the only observable change**, and only when registering a **brand-new subject** — array `items` goes from `["null", record]` to `record`. Existing subjects are left untouched (we only register if the subject doesn't already exist). Consumers of a *new* subject can't have been built against the old shape, so no compat break.
- **ClickHouse & Tinybird: no change.** Neither wraps a `Tuple` in `Nullable`, so both emit `Array(Tuple(...))` regardless of the element's nullability. (The `Array(Nullable(T))` vs `Array(T)` distinction only applies to a *scalar* element `T`, which these UDFs never produce — they always emit `List<Struct>`.)
- **Postgres:** arrays map to `JSONB`, which encodes no element name or nullability → no change.
- **Kafka-JSON and the message/JSON plugin sinks (s2, pubsub, eventbridge_partner):** serialize runtime rows, register no schema → no change. No MySQL sink exists.
- The element **rename to `"item"` is invisible to every sink** — arrays are positional; Avro drops the element name in canonical form; ClickHouse/Postgres don't name array elements.

Net: no data change anywhere; no change to any existing table/subject on upgrade; the only delta is a more-accurate (non-nullable) element declaration on a *newly-created* Avro subject.

xiaom (approved):

PR #24, opened 2026-06-17
feat: configurable Kafka sink compression codec (STRM-6290)

rslowinski:
## What

Adds an optional `compression` field to the Kafka sink, accepting `none`, `gzip`, or `lz4`. Defaults to `lz4`, which is the codec the producer has always used — so existing pipelines are unaffected.

```yaml
sinks:
  my_sink:
    type: kafka
    from: my_source
    topic: my.topic
    data_format: avro
    compression: gzip   # new; defaults to lz4 when omitted
```

## Why

The producer's compression codec was hardcoded to `lz4`. Some Kafka-compatible brokers don't accept it: Azure Event Hubs' Kafka endpoint, for example, supports only `gzip` or no compression and rejects lz4. There was no way to change the codec from pipeline config, so those brokers were unusable as a Kafka sink.

## How

- New `KafkaCompression` enum in `streamling-config` with a validating deserializer (rejects unknown codecs at config load with a clear message), mirroring the existing `ClickHouseCompression` pattern.
- Threaded through the topology `KafkaSink` → `KafkaSinkTableProvider` → `KafkaSink` connector.
- `compression.type` is set on the producer *after* the optimizer defaults, so the per-sink value takes precedence over the optimizer's `lz4`. Producer config assembly was extracted into `build_producer_config` so it can be unit-tested without opening a connection.

## Testing

- Unit: enum parsing (codecs, case-insensitivity, default, rejection); topology parsing; `build_producer_config` asserts the codec reaches the producer and overrides the optimizer default.
- E2E (`test_kafka_sink_gzip_compression`): produce → run with `compression: gzip` → consume, verifying gzip messages round-trip. All `test_kafka_sink_*` e2e tests pass locally against k3d Redpanda (6/6).
- `just fix && just lint` clean.

## Assumptions & unknowns (for reviewers)

- **Codec scope:** intentionally limited to `none`/`gzip`/`lz4`. These are always linked by librdkafka. `snappy`/`zstd` are omitted for now — `zstd` needs libzstd linked at build time, which I haven't verified. Easy to add later behind the same enum.
- **Not validated against a real Azure Event Hubs broker** — only local Redpanda. The compression mismatch is fixed, but EH has other constraints (e.g. message size limits, idempotence/`acks` behavior) that are out of scope here and unverified.

### DIFF
--- a/crates/streamling-connectors/src/table_providers/kafka.rs
+++ b/crates/streamling-connectors/src/table_providers/kafka.rs
@@ -2,7 +2,7 @@ mod config_optimizer;
 mod metadata;
 mod schema_registry;
 
-use streamling_config::KafkaConfig;
+use streamling_config::{KafkaCompression, KafkaConfig};
 use streamling_core::checkpoints::channels::{send, subscribe};
 use streamling_core::checkpoints::checkpoint_management::{
     CHECKPOINT_COORDINATOR_CHANNEL, CheckpointEpoch, CheckpointMessage,
@@ -1966,6 +1966,8 @@ pub struct KafkaSink {
     message_max_bytes: Option<u32>,
     /// Number of parallel producers (each with independent connections/queues)
     parallelism: usize,
+    /// Producer compression codec (maps to librdkafka's compression.type)
+    compression: KafkaCompression,
 }
 
 impl KafkaSink {
@@ -1983,6 +1985,7 @@ impl KafkaSink {
         batch_flush_interval_ms: Option<u64>,
         message_max_bytes: Option<u32>,
         parallelism: Option<usize>,
+        compression: KafkaCompression,
     ) -> Self {
         let subject_name_strategy = SubjectNameStrategy::TopicNameStrategy(topic.to_owned(), false);
         let topic_partitions = topic_partitions.unwrap_or(DEFAULT_NUM_PARTITIONS);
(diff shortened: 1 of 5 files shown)


crates/streamling-connectors/src/table_providers/kafka.rs
```diff
@@ -2034,48 +2038,69 @@ impl KafkaSink {
...
     }
 
+    /// Assemble the librdkafka producer configuration.
+    ///
+    /// Pure and network-free for non-WarpStream brokers, so it can be
+    /// unit-tested by inspecting the resulting `ClientConfig` without creating a
+    /// producer. Per-sink tunables are applied *after* the optimizer defaults so
+    /// they take precedence — in particular `compression.type`, which lets a
+    /// sink override the optimizer's default codec for brokers like Azure Event
+    /// Hubs that reject lz4.
```
jeffling:
```suggestion
```

crates/streamling-connectors/src/table_providers/kafka.rs
```diff
@@ -2703,6 +2733,77 @@ impl TableProvider for KafkaSinkTableProvider {
 mod tests {
     use super::*;
 
+    /// Minimal non-WarpStream config so `build_producer_config` stays
+    /// network-free (the WarpStream branch performs IMDS lookups).
```
jeffling:
```suggestion
```

jeffling (changes requested):
Remove downstream-specific comments

jeffling (approved):

PR #34, opened 2026-06-24
Experiment: Use Mimalloc3

jeffling:
Streamling dynamically loads plugin cdylibs over an abi_stable FFI boundary whose RVec/RString/RBox/RArc memory is allocated by the plugin and dropped by the host. That is only safe when host and plugins share one allocator heap.

Link mimalloc v3's single-translation-unit override (src/static.c built with -DMI_MALLOC_OVERRIDE) directly into the executable via build.rs, so the binary defines malloc/free/calloc/realloc/posix_memalign as exported symbols. With --export-dynamic (-Wl,-E) these land in the dynamic symbol table, and every dlopen'd plugin resolves its malloc to the same mimalloc heap as the host.

This deliberately does NOT use #[global_allocator]: that gives the host a private mimalloc heap while plugins keep libc malloc, so the host dropping a plugin-allocated object is a cross-heap free (verified to SIGSEGV). Only process-wide symbol interposition yields one shared heap.

Gated to target_os = "linux" (it is a glibc interposition mechanism); STREAMLING_NO_MIMALLOC=1 opts out to the platform default. mimalloc is MIT licensed and vendored under crates/streamling/vendor/mimalloc.

Verified: the release binary exports malloc==mi_malloc; a real plugin->print pipeline ran all cross-boundary batches + 17 checkpoints cleanly under mimalloc; a regression test asserts the override stays active.


jeffling:
image
Running on a few large pipelines showed similar improvements. Especially for pipelines that did a lot of unnesting of large objects, memory re-use seems to be greatly enhanced.

sfsf9797:
nice!!!

sap1ens (approved):

PR #39, opened 2026-06-30
feat(kafka): require a resolvable primary key for Kafka sinks (STRM-6281)

kashive:
## Summary

Closes STRM-6281.

Kafka was the only sink that could **crashloop without a usable primary key while still passing config validation**. ClickHouse is already protected because `primary_key` is a required field. This PR closes that gap for Kafka.

Two distinct bad states existed for a Kafka sink with no usable key:
- **`None`** → the sink silently drops to round-robin partitioning with no log compaction (a quiet behavior change).
- **`Some("")`** → crashloop at runtime: `primary key column '' not found in batch`.

### Where the empty string came from (answers the thread's open question)

The runtime `primary key column '' not found in batch` crashloop Jonathan reported is reachable when an **empty-string** primary key flows into a Kafka sink. The only producer of that empty string is the `goldsky-plugins` **abstract-dataset preprocessor**, which emits `primary_key: ""` for datasets that have no primary key in the CMS:

```rust
// streamling-goldsky-plugins/src/topology/dataset_preprocessor.rs
let pk = if pk.trim().is_empty() {
    warn!("Abstract dataset '{}' version '{}' has no primary key - using empty string", ...);
    String::new()
```

`PrimaryKeyMetadata::from_str` filters empty columns, so `""` became a `PrimaryKeyMetadata { columns: [] }` that passed schema validation and serialized back to `""` — reaching the sink as `Some("")` and crashing in `extract_keys_from_batch`.

### The fix (respects Yaroslav's Avro-discovery concern)

Rather than making `primary_key` a required YAML field (which would break the many topics whose key is discovered at runtime from the Avro schema), validation runs at **pipeline-build time** — *after* the source's Avro-schema discovery and *after* primary-key propagation. A key supplied by the sink config, an upstream node, or the source's Avro schema all satisfy it; only a genuinely missing/empty resolved key is rejected.

- New `PrimaryKeyRegistry::require_primary_key_for_sink(...)` requires a non-empty resolved key and returns a clear, user-facing error (also surfaced through `--validate`).
- The Kafka sink build path now uses it instead of `track_primary_key_for_transform_or_sink`.

Net effect: a runtime crashloop (or silent no-key behavior) becomes an up-front validation error that names the sink and tells the operator how to fix it.

## Test plan

- [x] Unit tests (`cargo test -p streamling --lib`, 66 passing) — written test-first and watched fail before the fix:
  - rejects empty-string key, blank/comma-only key, missing key with no upstream, and **propagated** empty key
  - accepts an explicit key and a key **propagated from a source** that discovered it from its Avro schema (`SchemaInferred`)
- [x] `just fix` + `just lint` clean (clippy `-D warnings`)
- [x] e2e test added (`test_kafka_sink_without_primary_key_fails_validation`) asserting a Kafka sink with no resolvable key fails `--validate`. Compiles under `cargo clippy --all-targets`; **not run locally** because the k3d cluster is down and kubectl is pointed at a prod context — relying on CI for the e2e run.

## Notes / follow-ups

- The empty-string origin (`goldsky-plugins` abstract-dataset preprocessor) is left as-is; this data-plane guard catches it regardless of source and is the most general fix. A follow-up could make that preprocessor fail loudly / omit the key instead of emitting `""`.
- This is intentionally consistent with the existing ClickHouse guard, but deferred to post-resolution so Avro/propagated keys still satisfy it.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/streamling-e2e/tests/validation.rs
+++ b/crates/streamling-e2e/tests/validation.rs
@@ -584,6 +584,13 @@ sinks:
         "Pipeline with an unresolvable Kafka sink primary key should be invalid, got: {:?}",
         validation
     );
+    // Attributed as a platform error (internal), since every dataset is expected to
+    // declare a primary key — so `--validate` reports success: false.
+    assert!(
+        !validation.success,
+        "A missing Kafka sink primary key is a platform (internal) error, so success should be false, got: {:?}",
+        validation
+    );
 
     let all_errors = validation.errors.join("\n");
     assert!(
--- a/crates/streamling-e2e/tests/validation.rs
+++ b/crates/streamling-e2e/tests/validation.rs
@@ -495,6 +495,105 @@ sinks:
 }
 
 // ============================================================================
+// STRM-6281: Kafka sink requires a resolvable primary key
+// ============================================================================
+
+/// A Kafka sink whose upstream has no primary key — none in the sink config, none on
+/// the source, and none discoverable from the Avro schema — must fail validation up
+/// front instead of crashlooping at runtime with `primary key column '' not found in
+/// batch`. This mirrors the abstract-dataset-with-no-CMS-primary-key scenario, where
+/// the preprocessor emits `primary_key: ""` that used to flow through to the sink.
+#[tokio::test]
+async fn test_kafka_sink_without_primary_key_fails_validation() {
+    init_tracing();
+
+    let ctx = TestContext::new()
+        .await
+        .expect("Failed to create test context");
+
+    // TEST_SCHEMA has no `doc` with primaryKeys, so nothing is discovered from Avro.
+    ctx.kafka
+        .register_schema(TEST_SCHEMA)
+        .await
+        .expect("Failed to register schema");
+
+    let records: Vec<TestRecord> = (0..5)
+        .map(|i| TestRecord {
+            block: i,
+            id: format!("id_{}", i),
+            data: format!("data{}", i),
+        })
+        .collect();
+
+    ctx.kafka
+        .produce_avro_records(&records)
+        .await
+        .expect("Failed to produce records");
+
+    let output_topic = ctx
+        .create_kafka_topic("output")
+        .await
+        .expect("Failed to create output topic");
+
+    // No primary_key on the source OR the sink, and no embedded Avro primary key:
+    // there is nothing to resolve or propagate to the Kafka sink.
+    let pipeline = format!(
+        r#"
+sources:
+  test_kafka_source:
+    type: kafka
+    topic: {input_topic}
+    starting_offsets: earliest
+
+transforms: {{}}
+
+sinks:
+  kafka_sink:
+    type: kafka
+    from: test_kafka_source
+    topic: {output_topic}
+    topic_partitions: 1
+    data_format: avro
+"#,
+        input_topic = ctx.kafka_topic,
+        output_topic = output_topic.topic,
+    );
+
+    let output = ctx
+        .run_pipeline_raw(
+            &pipeline,
+            PipelineOpts::new().record_limit(5).arg("--validate"),
+        )
+        .await
+        .expect("Failed to run pipeline");
+
+    assert!(
+        !output.status.success(),
+        "Kafka sink without a resolvable primary key should exit non-zero under --validate"
+    );
+
+    let validation: ValidationOutput = serde_json::from_str(&output.stdout).unwrap_or_else(|e| {
+        panic!(
+            "Failed to parse validation JSON from stdout: {}\nstdout was:\n{}\nstderr was:\n{}",
+            e, output.stdout, output.stderr
+        )
+    });
+
+    assert!(
+        !validation.is_valid,
+        "Pipeline with an unresolvable Kafka sink primary key should be invalid, got: {:?}",
+        validation
+    );
+
+    let all_errors = validation.errors.join("\n");
+    assert!(
+        all_errors.contains("kafka sink 'kafka_sink' requires a primary key"),
+        "Errors should mention the Kafka sink primary key requirement, got: {:?}",
+        validation.errors
+    );
+}
+
+// ============================================================================
 // STRM-5695: u256 comparison combined with boolean predicate via AND/OR
 // ============================================================================
 
--- a/crates/streamling/src/lib.rs
+++ b/crates/streamling/src/lib.rs
@@ -29,7 +29,7 @@ use streamling_core::operators::external_handlers::{ExternalHandlerConfig, Exter
 use streamling_core::operators::wasm_runner::WasmRunnerNode;
 use streamling_core::session::{SessionManager, SubqueryHandling};
 use streamling_core::topology::PipelineTopology;
-use streamling_core::{streamling_err, streamling_user_bail, streamling_user_err};
+use streamling_core::{streamling_bail, streamling_err, streamling_user_bail, streamling_user_err};
 use tokio::time::timeout;
 use tracing::{debug, error, info, warn};
 // Re-export for convenience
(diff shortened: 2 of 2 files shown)


kashive:
# QA Report — STRM-6281: Require a primary key for Kafka sinks

**Date:** 2026-07-01
**Verdict:** PASS — all 7 cases match expectations in both the binary-direct and control-plane endpoint phases.

## Feature under test

A Kafka sink without a usable primary key either silently drops to round-robin
partitioning (no log compaction) or crashloops at runtime with
`primary key column '' not found in batch` — the latter happening when an
upstream node propagates an empty-string key (as the abstract-dataset
preprocessor does, `primary_key: ""`, for datasets with no PK in the CMS).

The STRM-6281 change adds `PrimaryKeyRegistry::require_primary_key_for_sink`,
called from the Kafka sink build branch **after** Avro-schema discovery and
primary-key propagation. It requires a resolved, **non-empty** key and, when
none can be resolved, fails the pipeline build. The failure is raised as an
**internal (platform) error** (`streamling_bail!`), not a user error: every
dataset is expected to declare a primary key, so a missing one is a
platform/dataset invariant violation. Consequences:

- `streamling --validate` reports `success: false` (platform-attributed) in
  addition to `is_valid: false`.
- The error log is tagged `error.internal=true`, routing it to the platform team.
- The rule is Kafka-specific; other sinks (e.g. `blackhole`) are unaffected.

This is distinct from the pre-existing **user** error for a PK that names a
column not present in the schema, which stays `success: true` / `is_valid: false`.

## Environment / setup

| Item | Value |
|------|-------|
| streamling-1 branch | `avishek/strm-6281-require-primary-key-validation-for-kafka-sinks` |
| streamling-1 commit | `58bf7198` — "fix(kafka): attribute missing-primary-key as a platform (internal) error" |
| streamling-cloud-1 branch / commit | `feat/node-backpressure-metrics` / `095715e` |
| Toolchain | rustc 1.89.0 (aarch64-apple-darwin) |
| Validator binary | `target/debug/streamling` (debug build), `DYLD_LIBRARY_PATH` set to the toolchain lib dir |
| Agent run | `just run` (ENV=prod) → `streamling-agent/config.yaml`, bound `0.0.0.0:8124`, `allowed_projects: ["*"]`, `supervisor_enabled: false` |
| Kube context | `arn:aws:eks:us-west-2:301417190815:cluster/mercury-prod` (AWS SSO profile `prod`, `kubectl proxy --port=8001`) |
| Validator staging | `/opt/streamling/streamling-validator` → symlink to the debug binary (agent path is a hardcoded const) |
| Project id (URL) | `qa-strm-6281` (placeholder; `allowed_projects` is `*`) |
| Fixtures | `/tmp/strm6281-qa/` — `data.csv` (`id,value` + 2 rows), `case_{a..g}.yaml`, `case_{a..g}.request.json` |

**Scope note:** validate-only. No apply/deploy/pause/delete was ever called, and
the validate handler does not touch Kubernetes, so running against the prod
context is safe. The supervisor was disabled and auto-pause/delete are dry-run.

**Approach:** every case is an infra-free bounded **file source** (tiny local
CSV) feeding a **Kafka sink** (Kafka sinks do no network I/O during `--validate`,
so no broker/registry is needed). Two phases:

1. **Binary-direct** — `streamling --validate <case>.yaml`, inspecting the full
   JSON (`success`, `is_valid`, `errors`) to confirm the platform-error
   attribution (`success:false`).
2. **Endpoint** — `POST /pipeline/qa-strm-6281/validate` on the locally-run
   agent whose hardcoded validator is symlinked to the local build. The endpoint
   response (`ValidatePipelineResponse`) exposes only `valid`/`errors`/`warnings`
   — not `success` — so the platform attribution is verified in phase 1 and
   corroborated by the agent's WARN log in phase 2.

## Test matrix and results

Legend: **P** = platform/internal error (`success:false`), **U** = user error
(`success:true`), **V** = valid.

| Case | Pipeline | Expected | Binary `success` / `is_valid` | Endpoint `valid` | Result |
|------|----------|----------|-------------------------------|------------------|--------|
| A | kafka sink, no PK; source no PK | INVALID (P) | `false` / `false` | `false` | PASS |
| B | source `primary_key: ""`; sink no PK | INVALID (P) — empty-string propagation | `false` / `false` | `false` | PASS |
| C | sink `primary_key: "  ,  "` (blank) | INVALID (P) | `false` / `false` | `false` | PASS |
| D | sink `primary_key: id` (explicit) | VALID | `true` / `true` | `true` | PASS |
| E | source `primary_key: id`; sink omits (propagation) | VALID | `true` / `true` | `true` | PASS |
| F | `blackhole` sink from no-PK source (control) | VALID (rule is Kafka-specific) | `true` / `true` | `true` | PASS |
| G | sink `primary_key: does_not_exist` (contrast) | INVALID (U) | `true` / `false` | `false` | PASS |

### Per-case error/warning detail

- **A / B / C** — first error (identical across all three):
  `kafka sink 'kafka_sink' requires a primary key, but none could be resolved. Every dataset is expected to declare a primary key (in the sink config, an upstream node, the source's Avro schema, or — for abstract datasets — the CMS), so this indicates a platform/dataset configuration issue rather than a user error.`
  Binary tags the log `error.internal=true`; agent emits the WARN
  `Validator reported success=false (may be misclassified internal error)`.
- **D / E / F** — no errors. D and F carry the benign warning
  `Primary key not defined for source 'file_src'...`; E has no warnings (PK explicit on the source).
- **G** — first error (pre-existing **user** error, `success:true`):
  `Primary key validation failed for node 'kafka_sink': columns ["does_not_exist"] not found in schema. Available columns: ["id", "value", "_gs_op"]`

### Representative raw output (one per outcome)

Binary, Case A (platform-invalid):

```json
{
  "success": false,
  "is_valid": false,
  "errors": ["kafka sink 'kafka_sink' requires a primary key, but none could be resolved. ..."],
  "warnings": ["Primary key not defined for source 'file_src'. ...", "Cannot propagate primary key to 'kafka_sink': ..."]
}
```

Binary, Case G (user-invalid — note `success:true`):

```json
{
  "success": true,
  "is_valid": false,
  "errors": ["Primary key validation failed for node 'kafka_sink': columns [\"does_not_exist\"] not found in schema. Available columns: [\"id\", \"value\", \"_gs_op\"]"],
  "warnings": ["Primary key not defined for source 'file_src'. ..."]
}
```

Endpoint, Case A (invalid):

```json
{
  "valid": false,
  "errors": ["kafka sink 'kafka_sink' requires a primary key, but none could be resolved. ..."],
  "warnings": ["Primary key not defined for source 'file_src'. ...", "Cannot propagate primary key to 'kafka_sink': ..."]
}
```

Endpoint, Case E (valid):

```json
{ "valid": true, "errors": null, "warnings": null }
```

## Observations

- **Platform-vs-user attribution confirmed end-to-end.** Cases A/B/C are
  `success:false` (platform), and the agent emitted the
  `Validator reported success=false (may be misclassified internal error)` WARN
  for exactly those three. Case G — a genuine user misconfiguration — stayed
  `success:true` and did **not** trigger that WARN, while still being reported
  `valid:false`. The distinction the change introduces is observable in both
  phases.
- **Empty-string propagation (Case B) is caught.** An upstream `primary_key: ""`
  (the exact abstract-dataset preprocessor shape that caused the original
  runtime crashloop) is now rejected up front at validate time.
- **Blank/comma-only key (Case C) is caught.** `"  ,  "` resolves to zero
  columns and is rejected.
- **Rule is correctly scoped to Kafka (Case F).** A `blackhole` sink from the
  same no-PK source validates successfully — the new requirement does not
  over-reject other sink types.
- **Endpoint surface is narrower by design.** `ValidatePipelineResponse` returns
  only `valid`/`errors`/`warnings`; it does not expose `success`. Empty
  collections serialize as `null` at the endpoint vs `[]` from the binary — a
  cosmetic serialization difference, not a behavioral one. The platform
  attribution is preserved internally (WARN log) even though the response body
  can't express it.

## Follow-ups / bugs found

None. The feature behaves as designed across all cases in both phases. The only
minor note is the endpoint's inability to surface the platform-vs-user
`success` bit to callers; the agent already logs it as a WARN, so this is a
known, intentional limitation rather than a defect.

kashive:
👋

jeffling (approved):

PR #41, opened 2026-06-30
feat: add JSON support to the Kafka source

### LINKED ISSUE #40: Add JSON support for the Kafka source

sap1ens:
Closes #40 

Implements support for the JSON format for the Kafka source. Current syntax looks like this:

```
sources:
  raw_transactions:
    type: kafka
    data_format: json
    topic: test-json-topic
    primary_key: id
    schema:
      id: int32
      amount: int32
```

---

> [!NOTE]
> **Medium Risk**
> Touches the hot Kafka consumer decode path and batch/schema alignment; misconfiguration or schema mismatch fails the source at runtime, but Avro behavior is largely preserved behind the refactor.
> 
> **Overview**
> Adds **JSON** as an optional Kafka source payload format via `data_format: json` and a required `schema` map (column → Arrow type). Avro remains the default and unchanged; hybrid Kafka legs stay **Avro-only**.
> 
> The Kafka source refactors Avro-specific state into `KafkaSourceDecoding` / `KafkaSourceConverter`, decodes JSON with `JsonToArrowConverter` (including **projected** payload schemas so SQL column pruning stays aligned), and **rejects tombstone / empty payloads** for JSON with explicit errors. `_gs_op` still comes from `dbz.op` when missing from the payload.
> 
> Shared **`arrow_schema_from_type_map`** (with decimal parsing) replaces duplicated schema parsing in WASM and backs the JSON source schema. README documents JSON config; **e2e tests** cover decode and projection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ddd5df060ccd2c65973d8a30ae81331faf200b42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,8 @@ serialization. It's implemented as a custom DataFusion Table Provider (`TablePro
     - The size of the batches is controlled by the `record_batch_size` and `record_batch_interval_ms` parameters.
   - Adds operation type column (`_gs_op`) to track INSERT/UPDATE/DELETE operations (see `Upsert Semantics` section
     below). The operation type is determined by the `dbz.op` header value.
+  - Tombstone records (null/empty payloads, e.g. CDC deletes-as-tombstones) are **not supported** in any format
+    and fail the source. Represent deletes with a non-empty payload plus a `dbz.op=d` header instead.
 
 Kafka Source uses
 high-level [StreamConsumer](https://docs.rs/rdkafka/latest/rdkafka/consumer/struct.StreamConsumer.html) which handles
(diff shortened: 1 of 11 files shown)


kashive (approved):
Overall LGTM 🚀 

We might want to consider the PIPE-002 feedback from the AI.

Also, empty payload is considered internal err with `streamling_err!` but is more likely a user error since data is provided by the user. I guess in context of goldsky it's an internal but if context is streamling-only it is user.

sap1ens:
> We might want to consider the PIPE-002 feedback from the AI.
>
> Also, empty payload is considered internal err with streamling_err! but is more likely a user error since data is provided by the user. I guess in context of goldsky it's an internal but if context is streamling-only it is user.

Thanks! Addressed both concerns.

PR #45, opened 2026-07-07
fix: restart wedged broadcast via stall watchdog

kquintal:
Fixes a stuck consumer that stopped draining data, causing a freeze in the whole source with no error and no exit, so it needed a manual restart.

Here we add a **stall watchdog**, so stuck consumers get a generous window to self-heal; only a delivery still stuck past the deadline (5min) forces a `process::exit(1)` so the orchestrator restarts from the last checkpoint. This also covers the all-consumers-wedged case. Replaces the fail-fast approach from #44.

Should avoid pipelines from being stuck for more that 5mins before attempting a restart (configurable).

<details>
<summary><b>Detailed overview</b></summary>

## Problem

A shared source is scanned once and fanned out to N consumers (`BroadcastingExec` / `MultiSinkExec` -> `BroadcastStream`). The fan-out loop is serial: pull one batch, then deliver it to every consumer before pulling the next. Delivery retried a `Full` channel **forever**, giving up only on `Closed`.

When one consumer stopped draining its channel (a downstream sink stalled), that channel stayed full, its send retried forever, `join_all` never resolved, and the source was never polled again. The broadcast task is a detached `tokio::spawn` outside the top-level `select!`, so the wedge produced no error and no completion: the process stayed alive but frozen, cleared only by a manual restart.

## Approach

Track whether a batch is **mid-delivery** and watch it out-of-band:

- `in_flight: Option<Instant>` is set when a fan-out delivery starts and cleared when it completes. `Some(t)` means "a batch has been undeliverable since `t`"; `None` means the loop is idly waiting on the source. This lets the watchdog tell a genuine wedge apart from a source that simply has no new data (no false-positive restart while idle).
- A watchdog task polls `in_flight`. If a delivery has been stuck longer than `DEFAULT_STALL_DEADLINE` (5min), it forces a process exit so the orchestrator restarts from the last checkpoint.

```mermaid
graph TB
  RB[run_broadcast: set in_flight on delivery, clear on completion] -->|in_flight| WD[watchdog: poll every 5s]
  WD -->|in_flight stuck > 5min| EXIT[process::exit 1 -> restart from checkpoint]
  WD -->|in_flight None, idle| OK[do nothing]
  RB -->|consumer drains within window| RESUME[source resumes, no restart]
```

| Consumer state | Before | After |
|---|---|---|
| Full, self-heals within 5min | retry forever, whole source frozen | source pauses, then resumes; no restart |
| Full, never recovers | retry forever, silent hang | watchdog restart after 5min |
| All consumers wedged | silent hang | watchdog restart after 5min |
| Closed (receiver dropped) | warn every batch, never removed | pruned, warned once |

## Reviewer notes

- `DEFAULT_STALL_DEADLINE` (5min) is generous on purpose so a legitimately slow-but-recovering sink is never killed; it is a `const` with a `ponytail:` note to promote to config/env if needed. Open question: leave generous vs tighten toward checkpoint cadence.
- The restart is `std::process::exit(1)` from the watchdog task. It's abrupt (skips graceful checkpoint flush), but the pipeline is already wedged so there's nothing to flush; on restart it resumes from the last checkpoint. The exit action is injected into `run_watchdog` so it's unit-testable without killing the test process.
- Migrated this file's `consumers` mutex to `parking_lot` per repo rule (no guard is held across `.await`).
- Tradeoff accepted: while a single consumer is wedged, the shared source is paused for all consumers until it self-heals or the deadline fires. Keeping the others flowing would mean either dropping the stuck dataset (silent divergence) or unbounded per-consumer buffering (OOM). Neither is acceptable here.

## Test

- `watchdog_fires_when_delivery_stuck_past_deadline`: a delivery held in-flight past the deadline triggers the restart action.
- `watchdog_ignores_idle_source`: an idle loop (`in_flight = None`) never triggers a restart, guarding against false positives when the source simply has no data.

Ran `cargo fmt` and `cargo clippy -p streamling-core --all-targets` (clean with the repo's sanctioned `-A clippy::result_large_err`); all 11 broadcast tests pass. e2e not run locally (requires the k3s env, and forcing a consumer wedge end-to-end is impractical), hence draft.

</details>

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Claude Opus 4

---

> [!NOTE]
> **Medium Risk**
> Touches core multi-sink broadcast path and uses abrupt `process::exit(1)`, which can restart all datasets on a shared source after a long stall but avoids indefinite silent hangs.
> 
> **Overview**
> Adds a **stall watchdog** on `BroadcastStream` so a shared source that freezes when a consumer stops draining can recover without a manual restart.
> 
> While a batch is being fan-out delivered, `in_flight` records when delivery started; a background task polls every 5s and, if delivery stays blocked longer than **5 minutes** (`DEFAULT_STALL_DEADLINE`), logs an error and calls **`process::exit(1)`** so the orchestrator restarts from the last checkpoint. Idle sources (`in_flight` cleared) do not trigger restarts.
> 
> **Other behavior changes:** closed consumer channels are **pruned** from the consumer list instead of being retried and warned on every batch; `run_broadcast` sets **`stopped`** when the source ends so the watchdog exits. Consumer mutexes move to **`parking_lot::Mutex`**. Tests cover watchdog firing on stuck delivery and not firing when idle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5d4343522192b5ccaf20f3b07db98be699f51da3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sap1ens:
I'm not convinced this is the best solution to the problem. 

> When one consumer stopped draining its channel (a downstream sink stalled), that channel stayed full, its send retried forever, join_all never resolved, and the source was never polled again. The broadcast task is a detached tokio::spawn outside the top-level select!, so the wedge produced no error and no completion: the process stayed alive but frozen, cleared only by a manual restart.

This is the core of the backpressure design: if the sink is busy, do not emit any more data. I'm not sure if we can distinguish between stalled and slow/busy. If it was actually stalled, was there a bug in the sink implementation?

kquintal:
> I'm not convinced this is the best solution to the problem.
> 
> > When one consumer stopped draining its channel (a downstream sink stalled), that channel stayed full, its send retried forever, join_all never resolved, and the source was never polled again. The broadcast task is a detached tokio::spawn outside the top-level select!, so the wedge produced no error and no completion: the process stayed alive but frozen, cleared only by a manual restart.
> 
> This is the core of the backpressure design: if the sink is busy, do not emit any more data. I'm not sure if we can distinguish between stalled and slow/busy. If it was actually stalled, was there a bug in the sink implementation?

Yeah let's investigate at the sink itself, thats a better approach. Closing.

PR #50, opened 2026-07-09
fix: stop df54 ProjectionPushdown from deleting RebatchExec

rafadaguiar:
## Problem

After the DataFusion 49 → 54 upgrade (#22), high-throughput pipelines reading from plugin/dataset sources degraded severely — in the worst observed case to roughly a quarter of their previous throughput — with **zero error signal**: status, checkpoints, and logs all stayed healthy. Rolling the binary back restored full throughput immediately (with a large catch-up burst, proving the sources and sinks were never the bottleneck). Lower-rate pipelines on the same build were unaffected because they stayed under the reduced ceiling, which made the regression easy to miss. Observed sink batch sizes also changed on the new build — the tell that batching had been lost.

## Root cause

`RebatchExec` was a transparent alias for its input: it `delegate!`d `children()` **and the optimizer-negotiation hooks** (`try_swapping_with_projection`, `with_fetch`, `repartitioned`, `supports_limit_pushdown`) to `self.inner`. Each hook returns a rewritten subtree that does not contain the wrapper. DataFusion 49's projection pushdown was downcast-based and treated unknown nodes as barriers; **DataFusion 54 drives these trait hooks**, so `ProjectionPushdown` accepted the inner plan's swap and silently deleted the rebatcher from any plan with a projection above it — essentially all of them.

Postgres sinks depend on the rebatcher for accumulation (`postgres/mod.rs`: "accumulation is handled by RebatchExec"), so its loss turns each scan batch into a sink round-trip; the resulting backpressure throttles the whole synchronous pipeline. Script-transform rebatching is affected the same way (partially shielded by the WASM runner's internal accumulation).

Reproduced minimally: `Projection → Rebatch(1000) → source` + one `ProjectionPushdown` pass ⇒ `Projection → source`.

## Fix

Make `RebatchExec` a real plan node:
- `children()` exposes the input; `with_new_children` rewraps (arity-checked).
- Optimizer hooks fall back to the trait's no-op defaults → the node is a barrier.
- Row/order-preserving semantics stated explicitly (`maintains_input_order`, `cardinality_effect = Equal`, `partition_statistics` pass-through).
- `metrics()` no longer forwards the input's `MetricsSet` — the input is now a visible child, so forwarding would double-count.

## Testing

- `tests/rebatch_elision.rs` (new): fails on the old code (wrapper deleted, child hidden from traversal), passes with the fix; locks in barrier behavior + child visibility.
- `cargo test -p streamling-core`: 381 passed.
- fmt + clippy clean; `streamling`/`streamling-connectors` compile.

## Follow-up

`WrappingExec` delegates the same hook set to its inner plan and needs the same hardening — it carries the telemetry instrumentation (`output_rows`), so elision there would silently drop metrics. Addressed in #51 together with the other df54 planning regressions.

---

> [!NOTE]
> **Medium Risk**
> Changes physical plan shape and optimizer interaction on all sink/script rebatch paths; behavior fix is targeted but affects high-throughput batching correctness.
> 
> **Overview**
> **`RebatchExec` is no longer a transparent wrapper around its child.** The `delegate!` implementation that forwarded `children()` and physical-optimizer hooks (`try_swapping_with_projection`, `with_fetch`, `repartitioned`, etc.) to the inner plan is replaced with an explicit `ExecutionPlan` impl that exposes the real child, rewraps on `with_new_children`, and leaves hook methods at trait defaults so **ProjectionPushdown cannot swap the rebatcher out of the plan**.
> 
> Semantics are spelled out for optimizers: fixed name `RebatchExec`, order- and cardinality-preserving behavior, `benefits_from_input_partitioning = false`, and statistics pass-through—without forwarding inner **metrics** (avoids double-counting now that the child is visible).
> 
> A **regression test** runs `ProjectionPushdown` on a projection → `RebatchExec` → projection stack (sink-like shape) and asserts `RebatchExec` remains and both projections stay visible in the plan tree.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 09b73dabb83ee11a3bf531ba06e829434936acb4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


crates/streamling-core/tests/rebatch_elision.rs
```diff
@@ -0,0 +1,84 @@
+//! Regression check: DataFusion 54's ProjectionPushdown physical rule must not
```
jeffling: could this test be inlined in operators/rebatch.rs?

xiaom (commented):
have we reproduced it and tested this against any affected production pipeline?

crates/streamling-core/tests/rebatch_elision.rs
```diff
@@ -0,0 +1,85 @@
...
```
sap1ens: I'd move this test to `crates/streamling-core/src/operators/rebatch.rs`, we don't have any standalone test files in the streamling-core crate.

sap1ens (commented):
Thanks for the fix!

jeffling (approved):

PR #52, opened 2026-07-10
fix: plan source filters against the full payload schema, not the projected scan schema

rafadaguiar:
Third fix for the DataFusion 49 → 54 upgrade regressions (companions: #50, #51).

## Problem

A topology-level source `filter:` is defined against the source's **full payload schema**. Under DataFusion 54, scan projections get pushed through the surrounding extension nodes down to the Kafka source (they didn't under 49) — and a filter column is *typically not selected downstream*, since filtering is the only reason it's read. The source then prepared the filter against the already-projected exec schema and died at startup:

```
invalid filter for source '…'
Caused by: Schema error: No field named <col>. Valid fields are <projected columns…>
```

A perfectly valid filter becomes a fatal, crash-looping planning error, and the "Valid fields" list is misleading — it shows the projected schema, not the topic's schema, which makes the error look like a schema-evolution problem when it isn't.

## Fix

When a filter is present, build the source **unprojected**, plan the filter against the full schema, then re-apply the pushed projection above the filter. `StreamingProjectionExec` is used for the re-projection so batch metadata (checkpoint markers) survives, matching the existing `StreamingFilterExec`. The scan contract is preserved — output schema is still the projected one — and the no-filter fast path is unchanged.

Row cost is the same as pre-54 behavior: the filter needs those columns decoded anyway, and the projection still prunes immediately above it.

## Testing

- New unit test `source_filter_survives_scan_projection_pushdown`: reproduces the exact failure (filter referencing a projection-pruned column → "No field named c. Valid fields are a."), passes with the fix, and asserts the scan still returns the projected schema.
- `cargo test -p streamling-connectors`: 284 passed. fmt/clippy clean; `streamling` compiles.

---

> [!NOTE]
> **Medium Risk**
> Touches Kafka source physical planning and wrapper plan reordering on the hot path; behavior change is intentional for DF54 projection pushdown but could affect lag/event-time ordering if rebuild fails and falls back.
> 
> **Overview**
> Fixes a **DataFusion 54** regression where pushed scan projections could prune columns referenced by topology `filter:` on Kafka sources, causing fatal startup planning errors (`No field named …`).
> 
> **Kafka source planning** now builds the exec **without** the pushed projection when a source filter exists, compiles the filter against the **full** schema, then re-applies the engine’s column projection above the filter via **`StreamingProjectionExec`** (checkpoint metadata preserved, same pattern as **`StreamingFilterExec`**). Scan output schema stays projected; the no-filter path is unchanged.
> 
> **`WrappingSourceTableProvider`** peels **`StreamingProjectionExec` → `StreamingFilterExec`** so **`WrappingExec`** (side outputs, event-time) still sits **below** the filter, then rebuilds filter + projection on top.
> 
> Adds unit tests for projection+filter planning and wrapper reordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d93433b17a9cf5288e2175e5cd193243a7954b84. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/streamling-connectors/src/table_providers/kafka.rs
+++ b/crates/streamling-connectors/src/table_providers/kafka.rs
@@ -30,6 +30,7 @@ use datafusion::datasource::{TableProvider, TableType};
 use datafusion::error::Result;
 use datafusion::execution::TaskContext;
 use datafusion::logical_expr::Expr;
+use datafusion::physical_expr::expressions::col as physical_col;
 use datafusion::physical_expr::{EquivalenceProperties, PhysicalExpr};
 use datafusion::physical_plan::stream::RecordBatchReceiverStreamBuilder;
 use datafusion::physical_plan::{
@@ -38,6 +39,7 @@ use datafusion::physical_plan::{
     execution_plan::{Boundedness, EmissionType},
     filter::FilterExec,
     project_schema,
+    projection::ProjectionExec,
 };
 use streamling_core::formats::avro::{
     AvroToArrowConverter, FromArrowToAvroConverter, convert_avro_schema_to_arrow,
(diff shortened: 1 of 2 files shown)


xiaom:
lgtm

PR #53, opened 2026-07-10
fix(e2e): count all produced records in clickhouse update/delete record limits

rafadaguiar:
Fixes an intermittent failure in `test_append_only_mode_false_with_updates` (seen in CI: `id=1 should have the updated value / left: "original", right: "updated"`), plus the identical latent race in `test_append_only_mode_false`.

Both tests produce inserts and then a trailing update/delete in **separate producer calls**, but set `record_limit` (→ `STREAMLING__NUM_RECORDS_BEFORE_STOP`) to the **insert count only**. The ClickHouse sink counts written rows and stops the pipeline at the limit, so the outcome depends on Kafka consumer batch boundaries: if the trailing record lands in a later batch, the pipeline shuts down before it's written and the final-state assertion fails.

Fix: the limit counts every produced record. Deterministic regardless of batching.

Evidence it's a test bug, not a product bug: the same commit's functional twin passed 107/107 e2e minutes earlier; the failing/passing CI heads differed only in code comments.

---

> [!NOTE]
> **Low Risk**
> Test-only changes in e2e ClickHouse sink coverage; no production sink or runtime behavior is modified.
> 
> **Overview**
> Stabilizes **`test_append_only_mode_false_with_updates`** so the pipeline always processes every produced Kafka record before assertions run.
> 
> The test now sets the ClickHouse sink **`batch_size: 1`**, sets **`STREAMLING__RECORD_BATCH_SIZE`** to **`1`**, and raises **`record_limit`** to **inserts + updates** (3). Together this aligns stop-on-written-rows behavior with per-record batches: with **`append_only_mode: false`**, in-batch primary-key dedup meant the old insert-only limit could stop early depending on consumer batching, leaving **`id=1`** at **`original`** instead of **`updated`**.
> 
> A short comment documents why written-row limits are not deterministic without forcing one record per batch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4a1f5f648abbfc8617680a326bb214445c051d48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/streamling-e2e/tests/clickhouse_sink.rs
+++ b/crates/streamling-e2e/tests/clickhouse_sink.rs
@@ -774,7 +774,11 @@ sinks:
         topic = ctx.kafka_topic,
     );
 
-    let mut opts = PipelineOpts::new().record_limit(insert_records.len() as u64);
+    // Cover ALL produced records (inserts + the delete) — same batching race
+    // as test_append_only_mode_false_with_updates: with only the insert count
+    // the sink can stop before the delete arrives.
+    let mut opts =
+        PipelineOpts::new().record_limit((insert_records.len() + delete_records.len()) as u64);
     for (k, v) in clickhouse_env(&ctx) {
         opts = opts.env(&k, &v);
     }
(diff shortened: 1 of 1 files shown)


rafadaguiar:
Reworked after the first attempt timed out in CI: the sink's record limit counts *written* rows, and `append_only_mode: false` dedups by primary key within a batch — so raising the limit just inverted which Kafka batching schedule failed. The update test now forces one record per batch (`batch_size: 1` + record batch size 1) making the count deterministic, and the delete-test change is reverted entirely (deletes go via `ALTER TABLE DELETE` and never increment the written-row counter).

xiaom:
lgtm

PR #55, opened 2026-07-10
fix: make the live table schema authoritative for ClickHouse sink writes

rafadaguiar:
Fifth fix in the DataFusion-54 hardening series (#50–#53).

## Problem

The ClickHouse sink's write schema follows the pipeline's **input**, not the **table**. When the upstream schema is wider than a pre-existing table — e.g. a `SELECT *` transform over a source whose resolved schema has more columns than the table had at creation time — every write fails. Depending on the width difference it surfaces as:

- `Arrow error: number of columns(N+1) must match number of fields(N)` — `normalize_batch_for_clickhouse` paired columns **positionally** via a triple-zip and rebuilt the batch with every incoming column; or
- `NO_SUCH_COLUMN_IN_TABLE` from ClickHouse on the INSERT, retried forever — a wedged pipeline that still reports healthy liveness.

The positional zip also silently truncates when widths diverge the other way, which can **mislabel columns** — arguably worse than the crash.

## Fix

1. `write_all` fetches the live table's columns (new async `fetch_table_column_names`; the existing `fetch_schema` uses `block_on` and can't be called from async) and filters the write schema to real table columns, warning with the dropped names — the warning also serves as the diagnostic for *why* a batch was wider. `_gs_op` is kept for the CDC split and stripped before INSERT, as before. On listing failure, previous behavior is preserved.
2. `normalize_batch_for_clickhouse` pairs columns **by name** against the sink schema, mirroring the read-side `normalize_batch_from_clickhouse` (which was already fixed this way).

## Testing

- New e2e test `test_select_except_same_name_readd`: pre-creates the sink table narrower than the pipeline's output; on main it wedges (INSERT retries until the run times out), with the fix it passes (extra column dropped + warned, rows written, dedup semantics intact).
- `cargo test -p streamling-connectors`: 109 passed; core suite green; fmt/clippy clean.
- Full local e2e blocked by an environment issue in an unrelated Postgres/metrics test (fails identically on a pre-change build); relying on CI for the full suite.

## Follow-up

The Postgres sink write path deserves the same live-table-schema audit.

---

> [!NOTE]
> **Medium Risk**
> Changes core ClickHouse sink write and batch normalization on every INSERT; mitigated by fallbacks on column-list failure and an e2e regression for schema-wider-than-table cases.
> 
> **Overview**
> Fixes ClickHouse sink failures when the pipeline schema is **wider** than an existing table (e.g. upstream `SELECT *` after schema evolution): writes now follow the **live table**, not the declared input schema.
> 
> In **`write_all`**, after `CREATE TABLE IF NOT EXISTS`, the sink loads column names via new async **`fetch_table_column_names`** (`system.columns`, with error-body detection like DDL). The write schema is filtered to columns that exist in ClickHouse (keeping **`_gs_op`** for CDC); extras are **warned and dropped**. Empty listings or fetch errors fall back to the previous declared schema so writes do not wedge on empty INSERTs.
> 
> **`normalize_batch_for_clickhouse`** now maps columns **by name** to the sink schema (matching read-side **`normalize_batch_from_clickhouse`**): extra batch columns are dropped, missing sink columns error, and reordered batches no longer mislabel data. The old positional zip caused Arrow column-count mismatches or silent wrong-column INSERTs.
> 
> Adds e2e **`test_select_except_same_name_readd`**: narrower pre-created table, `SELECT * EXCEPT` + re-added `_gs_op`, verifies rows land with correct `id`/`value`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ffbac2dc947f75565547451af04722dc86b3bba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/streamling-connectors/src/table_providers/clickhouse.rs
+++ b/crates/streamling-connectors/src/table_providers/clickhouse.rs
@@ -2242,8 +2242,7 @@ impl ClickHouseClient {
                 "ClickHouse returned an error body while listing columns of '{}': {}",
                 table_name,
                 body.lines().next().unwrap_or_default()
-            )
-            .into());
+            ));
         }
         Ok(body
             .lines()
--- a/crates/streamling-connectors/src/table_providers/clickhouse.rs
+++ b/crates/streamling-connectors/src/table_providers/clickhouse.rs
@@ -969,6 +969,18 @@ impl DataSink for ClickHouseSinkExec {
         // ClickHouse and every INSERT naming them fails. `_gs_op` is kept —
         // the CDC split below needs it, and it is stripped before INSERT.
         let normalized_schema = match client.fetch_table_column_names(&self.table_name).await {
+            // An empty listing right after CREATE TABLE IF NOT EXISTS succeeded
+            // means the lookup is unreliable (database/name resolution mismatch,
+            // system.columns lag) — treat it like a failed fetch rather than
+            // dropping every column and issuing empty INSERTs.
+            Ok(live_columns) if live_columns.is_empty() => {
+                warn!(
+                    "[{}] ClickHouse column listing for table '{}' came back empty; \
+                     writing with the sink's declared schema",
+                    node_label, self.table_name
+                );
+                self.schema.clone()
+            }
             Ok(live_columns) => {
                 let live: std::collections::HashSet<&str> =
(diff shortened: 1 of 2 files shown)


xiaom:
lgtm

PR #56, opened 2026-07-11
fix: only append the enriched op column when the scan projection kept it

rafadaguiar:
Sixth fix in the DataFusion-54 hardening series (#50–#53, #55).

## Problem

With op enrichment enabled, the Kafka source appends the synthesized `_gs_op` column to every batch **unconditionally**, while building the batch against its **projected** schema. Under DataFusion 54, scan projections reach the source (they didn't under 49) — and a transform that never *reads* `_gs_op` (typical shape: `SELECT * EXCEPT (…, _gs_op), CASE … AS _gs_op` recomputing it from another column) lets the engine correctly prune `_gs_op` from the scan. Result: every batch is one column wider than its schema, and the pipeline crash-loops at the source with:

```
Arrow error: Invalid argument error: number of columns(9) must match number of fields(8) in schema
```

The error surfaces wrapped in "Sink future […] completed with error", which makes it look like a sink problem — it isn't; a diagnostic backtrace placed it in `KafkaSourceExec::execute`. Pipelines whose SQL references `_gs_op` keep it in the projection and are unaffected, which is why this shape is rare but deterministic.

## Fix

Compute once, at stream setup, whether the projected schema contains the op column; gate both the per-record row-kind collection and the column append on it. The `kafka_consumer.row_kind_count` metric is recorded regardless, as before.

## Testing

- New e2e `test_wide_source_except_precreated_versioned_table`: 9-field source, 4-column `EXCEPT` with a rename and a recomputed same-name `_gs_op`, pre-created two-argument `ReplacingMergeTree(version, is_deleted)` table. Fails on main with the width error; passes with the fix.
- All 10 clickhouse_sink e2e tests pass locally; `cargo test -p streamling-connectors` 284 passed; `just lint` clean.

## Follow-up (deliberately not here)

`include_metadata` columns are appended by the same assembly and would interact identically with a projection that prunes them; needs its own reproduction and change.

---

> [!NOTE]
> **Medium Risk**
> Touches core Kafka source batch assembly on the DataFusion 54 projection path; behavior change is narrow (only when `_gs_op` is pruned from the scan) and is covered by a new e2e test.
> 
> **Overview**
> Fixes Kafka source batches being **one column wider than the projected schema** under DataFusion 54 when op enrichment is on but the pushed-down scan **drops** `_gs_op` (e.g. `SELECT * EXCEPT (…, _gs_op), … AS _gs_op`).
> 
> At stream setup the source now sets `append_op_column` from `full_schema_projected` containing `_gs_op`, and uses that to gate collecting row-kind strings and appending the synthesized column. **`kafka_consumer.row_kind_count` still records on every message** when enrichment is enabled.
> 
> Adds e2e **`test_wide_source_except_precreated_versioned_table`**: wide Avro source, multi-column `EXCEPT` with rename and recomputed `_gs_op`, ClickHouse `ReplacingMergeTree` sink — the shape that previously failed with column/field count mismatch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3bcfc87aab7de5aab7da65a4bf92c8551a1fb83. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


xiaom:
lgtm

PR #57, opened 2026-07-12
feat: add zstd and lz4 compression to the ClickHouse sink, default to zstd

jeffling:
## What

Adds `zstd` and `lz4` as wire-compression options for ClickHouse sink INSERTs and makes `zstd` the default (previously no compression).

- `ClickHouseCompression` gains `Zstd` (default) and `Lz4` variants, accepted case-insensitively in config and pipeline YAML alongside `none`/`gzip`.
- `send_arrow_batch` encodes with `zstd::stream::write::Encoder` (default level 3) or `lz4_flex::frame::FrameEncoder` inside the existing spawn_blocking, wrapping the Arrow IPC writer directly so there is no extra copy, and sets the matching `Content-Encoding` header.
- `compression_level` remains gzip-only; lz4 frame format has no level knob.
- `zstd` and `lz4_flex` were already in the lockfile transitively, so no new crates enter the dependency tree.

## Testing

- Config unit tests cover parsing all four codecs and the new zstd default.
- Mockito tests assert the `Content-Encoding: zstd` / `lz4` headers on outbound requests.
- E2e suite now covers all three codecs against real ClickHouse: `test_basic_kafka_to_clickhouse` uses gzip, `test_multiple_batches` uses lz4, and the remaining tests exercise the zstd default. Full clickhouse_sink suite run locally: 9/9 passed, including update/delete/dedup row-content assertions.
- `just lint` clean.

## Upgrade note

The default `compression` changes from `none` to `zstd`. Deployments that don't set `compression:` will start sending zstd-compressed INSERT bodies (`Content-Encoding: zstd`) on upgrade. ClickHouse has accepted this since ~21.x and ClickHouse Cloud always supports it; set `compression: none` explicitly to keep the old wire behavior (e.g. for proxies that inspect POST bodies).

## Benchmarks

We also experimented with a highly optimized native-protocol client (`clickhouse-c-rs`, C library bindings speaking ClickHouse's native TCP protocol) and with brotli, on the `clickhouse-c-rs-github-arrow` experiment branch. Full write-up in that branch's `crates/streamling-connectors/tests/COMPRESSION_BENCHMARK.md`. Setup: AMD EPYC-Genoa, ClickHouse 24.8.14 in k3d, `id Int64, value String, timestamp Int64` MergeTree, median of 5 runs.

### INSERT throughput

| Batch | HTTP none | HTTP gzip(6) | HTTP lz4 | HTTP zstd(3) | HTTP brotli(6) | Native LZ4 | Native zstd |
|---|---|---|---|---|---|---|---|
| 1K | 2.8ms | 3.3ms | 2.2ms | 2.6ms | 4.8ms | 2.2ms | 3.3ms |
| 10K | 6.1ms | 10.5ms | 5.4ms | 5.5ms | 13.9ms | 5.4ms | 5.9ms |
| 100K | 38.7ms | 86.8ms | 32.9ms | 36.0ms | 186.4ms | 30.5ms | 33.7ms |
| 500K | 196.4ms | 494.8ms | 171.5ms | 136.1ms | 922.0ms | 120.4ms | 126.4ms |
| 1M | 344.7ms | 934.4ms | 277.7ms | 284.5ms | 1797.8ms | 258.4ms | 287.5ms |

Speedup vs gzip (the previous compressed option): lz4 1.5–3.4x, zstd 1.2–3.6x, native LZ4 1.5–4.1x, growing with batch size. Brotli is 0.5–0.75x (slower than gzip).

### Wire payload (100K rows, varied ~20B strings — realistic)

| Codec | Compressed size | Ratio |
|---|---|---|
| gzip(6) | 832.4 KB | 5.0x |
| lz4 (frame) | 1681.9 KB | 2.5x |
| zstd(3) | 572.9 KB | 7.2x |
| brotli(6) | 419.8 KB | 9.8x |
| Native LZ4 | 1687.1 KB | 2.7x |

### Why this PR ships HTTP zstd/lz4 and not the native path or brotli

- **zstd is the sweet spot**: 3–4x faster than gzip with ~35% smaller payloads, low CPU cost, no new dependency tree entries — hence the new default.
- **lz4** is the fastest HTTP option for small batches (1K–10K rows), at ~3x larger payloads than zstd.
- **Native LZ4** is the fastest path overall at scale (an extra 1.1–1.5x over HTTP zstd at 100K+ rows, mostly from skipping Arrow IPC encoding), but it costs a C build dependency (`liblz4-dev`/`libzstd-dev`), a git-pinned crate, and a ~900-line Arrow→native-block conversion layer, and it sends ~3x more bytes over the wire, which flips the advantage on WAN links. Not worth it for now; the write-up preserves the details if we revisit.
- **Brotli** compresses best (up to 9.8x on realistic data) but is CPU-bound and slower than gzip end-to-end — only interesting for bandwidth-limited links, so it was left out.

Caveats: localhost benchmarks (network bandwidth zeroed out, so smaller payloads don't help throughput here — over a WAN the ranking shifts toward zstd/brotli), and zstd was measured at level 3, the same fixed level this PR uses.


jeffling:
Addressed the review findings:

- **CONV-004 (no e2e for the default zstd path):** the default path is already covered — 7 of the 9 tests in `clickhouse_sink.rs` omit `compression:` and therefore run the zstd default against real ClickHouse (all 9 passed locally, including update/delete/dedup row-content assertions). Added a doc comment on `test_deduplication` making that explicit.
- **Default-change callout:** added an "Upgrade note" section to the PR description covering the `none` → `zstd` default shift, minimum ClickHouse version, and how to opt out with `compression: none`.

sap1ens (approved):
Nice!

PR #58, opened 2026-07-13
fix: forward checkpoint markers through unnest without waiting for data

rafadaguiar:
## Problem

A pipeline with an unnest downstream of a highly selective filter stops checkpointing entirely whenever the filter matches no rows after startup. Data keeps flowing, but the first checkpoint epoch never finalizes: the sink never receives the marker, the epoch times out after 300s and keeps retrying, and no checkpoint state is ever persisted — so any restart rewinds to the last checkpoint from before the stall.

## Root cause

`UnnestStream` queued checkpoint markers that arrived before any non-empty batch (`pending_checkpoints`), releasing them only with the first non-empty batch or at end of stream. On a live stream neither is guaranteed to happen: the checkpoint producer does not start a new epoch until the in-flight one resolves, so once the first marker is held, nothing else ever arrives to flush it. The producer and the unnest wait on each other forever.

The queue existed to keep empty-batch schemas consistent with resolved (possibly type-promoted) schemas, but holding markers cannot deliver that: `safe_take`'s Utf8 → LargeUtf8 promotion is per-batch, so consumers already observe mixed types across non-empty batches whenever promotion kicks in mid-stream. The end-of-stream path also already emitted marker-only batches with the static plan schema as a fallback.

## Fix

Forward markers immediately as empty batches via the existing `create_empty_checkpoint_batch` helper (resolved schema when known, static plan schema otherwise). Remove the `pending_checkpoints` machinery.

## Tests

- Unit: `test_unnest_forwards_marker_immediately_on_open_stream` — a marker-carrying empty batch followed by a stream that stays open must yield the marker batch promptly; hangs (5s timeout) on the previous code.
- E2E: `test_checkpoint_progresses_when_unnest_yields_no_rows` — full pipeline whose filter matches nothing ahead of an unnest; asserts checkpoint state is persisted while it runs.
- Existing unnest checkpoint-metadata tests unchanged and passing.

---

> [!NOTE]
> **Medium Risk**
> Changes checkpoint propagation in a core streaming operator; behavior is narrower (immediate forward vs hold) and covered by new unit and e2e tests, but incorrect marker handling could still affect pipeline recovery.
> 
> **Overview**
> Fixes **checkpoint deadlock** when `UnnestStream` sits behind a filter that drops every row: markers were queued in `pending_checkpoints` until the first non-empty batch or EOS, but the checkpoint producer will not advance epochs until the in-flight marker is acknowledged—so on a live stream that never unnests rows, markers never flushed and checkpointing stalled indefinitely.
> 
> **`UnnestStream`** now emits checkpoint-carrying empty batches **immediately** via `create_empty_checkpoint_batch` when `build_batch` returns no rows (resolved schema when known, static plan schema otherwise). The `pending_checkpoints` queue, merge-into-non-empty-batch logic, and end-of-stream flush are removed.
> 
> Adds **`test_unnest_forwards_marker_immediately_on_open_stream`** (marker batch + indefinitely open input must yield within 5s) and e2e **`test_checkpoint_progresses_when_unnest_yields_no_rows`** (Kafka → impossible filter → unnest → Postgres; asserts checkpoint rows in state table while the sink gets no data rows).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1eb3e9462ec700e2acfbdddaafcb708931a3227e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/streamling-core/src/operators/unnest.rs
+++ b/crates/streamling-core/src/operators/unnest.rs
@@ -202,7 +202,6 @@ impl ExecutionPlan for StreamingUnnestExec {
             options: self.options.clone(),
             metrics,
             resolved_schema: None,
-            pending_checkpoints: Vec::new(),
         }))
     }
 
@@ -259,10 +258,6 @@ struct UnnestStream {
     /// type promotion occurred (e.g., Utf8 -> LargeUtf8 via safe_take fallback).
     /// Used to ensure empty batches use consistent types with non-empty batches.
     resolved_schema: Option<SchemaRef>,
-    /// Pending checkpoint messages from early empty batches.
-    /// These are queued until we have a resolved schema to ensure schema consistency.
-    /// They will be emitted with the first non-empty batch or when the stream ends.
-    pending_checkpoints: Vec<CheckpointMessage>,
 }
 
 impl RecordBatchStream for UnnestStream {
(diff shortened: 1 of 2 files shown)


xiaom:
lgtm

PR #65, opened 2026-07-21
fix: recover PostgreSQL sinks from read-only failover (SQLSTATE 25006)

xiaom:
After a primary→replica failover, pooled connections keep pointing at a read-only server and every write fails with SQLSTATE 25006 until the process restarts. Gate idle-connection reuse with a before_acquire hook that runs SHOW transaction_read_only and evicts read-only sessions, so the pool reconnects and sinks resume as soon as a writable primary is reachable — no new retry machinery needed; the existing retry-forever loop in the sink already covers the failure window.

The SHOW round-trip replaces the test_before_acquire ping (any query proves liveness), so idle-acquire cost is unchanged.

The e2e test produces records in two waves (the second only after the database is read-only) so it cannot pass without exercising the window, and deliberately does not terminate backends when the failover resolves — recovery must come from the eviction gate itself.

---

> [!NOTE]
> **Medium Risk**
> Changes shared PostgreSQL pool acquisition for all sinks and pg aggregation; mis-eviction or query failures could affect connection reuse, but behavior is narrowly scoped to read-only detection and existing sink retries still apply.
> 
> **Overview**
> PostgreSQL sink pools no longer reuse connections that are in read-only mode after a primary→replica failover. **`test_before_acquire`** is turned off and replaced with a **`before_acquire`** hook that runs **`SHOW transaction_read_only`** and drops connections unless the value is **`off`**, forcing reconnects to a writable primary while still using that round-trip as the liveness check.
> 
> E2E coverage adds **`admin_execute`** on the cluster admin pool for **`ALTER DATABASE ... default_transaction_read_only`** and **`pg_terminate_backend`**, plus **`test_postgres_sink_read_only_failover`**, which runs the sink mid-flight through a simulated read-only window (two Kafka waves, backends killed only when entering read-only, not on recovery) and asserts all 50 rows land without restarting the process.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d983a69b20fb966d3f93ea7fae31b52dd6ae389b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/streamling-core/src/utils/pg.rs
+++ b/crates/streamling-core/src/utils/pg.rs
@@ -60,7 +60,22 @@ impl PostgresConnection {
             .acquire_timeout(Duration::from_secs(config.pool_acquire_timeout_secs))
             .idle_timeout(Duration::from_secs(config.pool_idle_timeout_secs))
             .max_lifetime(Duration::from_secs(config.pool_max_lifetime_secs))
-            .test_before_acquire(true)
+            .test_before_acquire(false)
+            // Discard idle connections that can no longer accept writes. After a
+            // primary→replica failover, pooled connections keep pointing at a
+            // read-only server and every write fails with SQLSTATE 25006 until
+            // the process restarts; evicting them here forces the pool to
+            // reconnect, so sinks resume as soon as a writable primary is
+            // reachable again. The SHOW round-trip also doubles as the liveness
+            // ping `test_before_acquire` used to perform.
+            .before_acquire(|conn, _meta| {
+                Box::pin(async move {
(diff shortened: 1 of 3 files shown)


jeffling (approved):

jeffling (commented):
does this require admin permissions to succeed?  that could be a blocker.

xiaom:
`SHOW transaction_read_only`  does not require admin permission.  The other admin permission is for testing setup.
will test it with a production pipeline and verify the failover handling actually works.

jeffling:
sounds good, ship it

PR #68, opened 2026-07-22
feat: add opt-in Postgres dynamic table cache

jeffling:
## Summary

- Add an opt-in in-memory membership cache for PostgreSQL dynamic tables.
- Enable caching only when the global `cache_enabled` flag is true and the table explicitly declares a timestamp-compatible `time_column`; legacy tables keep the existing batched lookup path.
- Populate the cache lazily through a repeatable-read PostgreSQL cursor, fetching 1,000 rows per page into `HashSet<Box<str>>` instead of building one `ARRAY_AGG` result.
- Check `MAX(time_column)` on every `dynamic_table_check` invocation. When the watermark advances, fetch only newer rows and extend the existing cache in place.
- Avoid cloning Arrow input strings on cached lookups and deduplicate values sent through the uncached lookup path.
- Emit structured logs for cache configuration, initial population, and incremental refreshes.

## Why

Repeated PostgreSQL membership queries are expensive for frequently evaluated dynamic-table filters. Keeping the membership set in process makes each cached lookup a direct hash-set probe, while cursor pagination bounds temporary load memory for large tables.

## Configuration

Caching remains disabled by default:

```yaml
dynamic_table_backend:
  postgres:
    cache_enabled: true
```

Each cached table must also opt in with a time column:

```yaml
transforms:
  membership:
    type: dynamic_table
    backend_type: Postgres
    backend_entity_name: persistent_data
    time_column: updated_at
```

The global flag can alternatively be set with `STREAMLING__DYNAMIC_TABLE_BACKEND__POSTGRES__CACHE_ENABLED=true`.

## Correctness and operational behavior

- Cached tables are append-only. Updates and deletes are not reflected in this mode.
- Every writer must take the same per-table PostgreSQL advisory transaction lock and assign `CLOCK_TIMESTAMP()` after acquiring it. Streamling's cached writer path follows this protocol.
- `time_column` must advance for every append and should be indexed so the watermark and incremental range queries stay cheap.
- Initial population occurs on the first cached lookup. Cached lookups wait while initial population or an incremental refresh is in progress; there is no fallback to the uncached query path during a load.
- Initial population and refresh logs include the table, entries, cursor pages, duration, and watermark. Refresh details are available at `debug` level.

## Validation

- `cargo test -p streamling-core dynamic_table::postgres::tests`
- `just e2e-test test_postgres_dynamic_table_cache_loads_and_refreshes_after_append`
- `just e2e-test test_postgres_dynamic_table_cache_refreshes_with_serialized_writers`
- `just e2e-test test_postgres_dynamic_table_cache_non_orderable_time_column_fails_fast`
- `just fix`
- `just lint`
- GitHub Actions `validate-and-test` and `e2e-tests`

### DIFF
--- a/README.md
+++ b/README.md
@@ -985,9 +985,9 @@ transforms:
 ```
 
 The cache is off by default and is used only when both settings are present. The initial lookup
-loads the full table. Each later `dynamic_table_check` batch reads `MAX(time_column)` and appends
-only rows newer than the cached maximum. Index the time column so these checks and range reads stay
-cheap.
+loads the full table through bounded PostgreSQL cursor pages. Each later `dynamic_table_check`
+batch reads `MAX(time_column)` and appends only rows newer than the cached maximum. Index the time
+column so these checks and range reads stay cheap.
 
 PostgreSQL dynamic tables are append-only when the in-memory cache is enabled. Updating or deleting
 existing membership, including removals, is not supported in this mode. Keeping `time_column` current
--- a/README.md
+++ b/README.md
@@ -985,9 +985,39 @@ transforms:
 ```
 
 The cache is off by default and is used only when both settings are present. Each
-`dynamic_table_check` batch reads `MAX(time_column)` and reloads the cache only when it changes.
-Index the time column so this check stays cheap. Tables without a reliable update timestamp, or
-deployments with `cache_enabled: false`, continue to use batched PostgreSQL lookups.
+`dynamic_table_check` batch reads only `MAX(time_column)` and reloads the cache when it changes.
+Index the time column so this check stays cheap.
+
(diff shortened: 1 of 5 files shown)


kashive (approved):
Overall LGTM! In-memory lookup for dynamic table checks 🙌  

Any memory concerns? Should we add some guard in regards to how many entries exist in the table at the first full-load? Or even when the watermark changes.

Also, do we have use case of deleting rows in the dynamic table? In this case, the cache would be stale right?

jeffling:
Unfortunately this doesn't support deletes yet. but I think we can turn the lookup table to a cdc-style table with an 'op' column next version

PR #72, opened 2026-07-24
fix: preserve checkpoint markers through multi-sink transform chains

rafadaguiar:
## Problem

Checkpoint markers travel as per-batch schema metadata. Streamling normally protects them by rewriting stock DataFusion operators (`ProjectionExec`/`FilterExec`/`UnnestExec`) into metadata-preserving `Streaming*` variants via physical-optimizer rules — but those rules traverse the plan tree through `children()`, and the per-sink plans built by the MultiSink extension planner live in `MultiSinkExec::sinks`, which `children()` does not expose. They are never rewritten.

A sink whose `insert_into` adds a stock operator — ClickHouse always interposes a `ProjectionExec`; Postgres does when the schema has u256/i256 or nested columns — then rebuilds every batch with its plan-time schema, silently dropping the markers. The sink keeps writing data but never acks a checkpoint epoch, and the coordinator stalls forever:

```
Checkpoint producer still waiting for finalization (...): epoch=1,
state=InProgress { acked_sinks: {"webhook_sink"}, ... } — missing sinks: ["clickhouse_sink"]
```

This hits any topology where such a sink shares a source/transform with another sink (which is what engages the multi-sink broadcast). The sibling sink (e.g. webhook — no added nodes) acks fine, which is why delivery-focused tests never caught it.

## Fix

`rewrite_sink_transform_chain()` rewrites each sink plan's transform chain into the `Streaming*` operators at planning time, mirroring what the optimizer rules do for the main tree. Traversal stops at the shared broadcast input (`TreeNodeRecursion::Jump`): that subtree is optimized in the main plan tree and is replaced by the broadcast stream at execute time, and not descending into it keeps the `Arc::ptr_eq` transform detection intact for sinks that add no nodes.

`StreamingProjectionExec::with_new_children` returns the streaming variant, so the rewrite also survives the runtime child-swap in `MultiSinkExec::execute`.

## Tests

- 3 unit tests for the rewrite: sink-side projection rewritten with the shared input forwarded untouched; identity (same `Arc`) when the sink adds no nodes; no descent into the shared input subtree.
- New e2e `test_multi_sink_checkpoint_finalizes_with_clickhouse_and_webhook`: ClickHouse + webhook sharing one Kafka source, 1s checkpoint interval, records produced in two waves so the run provably spans a full epoch cycle. Asserts a **positive** "Epoch finalized" (which requires the sink behind the projection to ack), no stall warnings, and delivery on both branches. The positive assertion matters: the stall warning only appears after 10s of waiting, so short runs would otherwise pass vacuously.
- Verified the e2e test **fails without the fix** (missing "Epoch finalized") and passes with it; full `multi_sink` (4/4) and checkpoint-related (7/7) e2e suites pass, `streamling-core` unit tests 392/392.

---

> [!NOTE]
> **Medium Risk**
> Touches multi-sink physical planning and checkpoint ack paths for transform-heavy sinks; behavior is narrowly scoped with targeted unit and e2e coverage.
> 
> **Overview**
> Fixes **checkpoint stalls** when multiple sinks share a source and one sink’s `insert_into` adds stock DataFusion operators (e.g. ClickHouse’s `ProjectionExec`) that were never rewritten by the physical optimizer.
> 
> Per-sink plans live under `MultiSinkExec::sinks`, outside `children()`, so checkpoint marker metadata on batch schemas was dropped in those transforms—the sink kept writing but never acked epochs while siblings (e.g. webhook) could ack, leaving the coordinator on “missing sinks”.
> 
> Adds **`rewrite_sink_transform_chain`** at multi-sink planning time to swap `ProjectionExec` / `FilterExec` / `UnnestExec` for metadata-preserving `Streaming*` variants, stopping at the shared broadcast input so transform detection and main-tree optimization stay unchanged. **Unit tests** cover rewrite, identity, and no descent into the shared subtree. **E2e** `test_multi_sink_checkpoint_finalizes_with_clickhouse_and_webhook` asserts epoch finalization (not just delivery) for ClickHouse + webhook with 1s checkpoints.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e99921fa302899312e1f2dded20e231bd65477e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/streamling-core/src/operators/broadcast.rs
+++ b/crates/streamling-core/src/operators/broadcast.rs
@@ -5,11 +5,15 @@
 pub(crate) mod broadcast_stream;
 
 use crate::operators::broadcast::broadcast_stream::BroadcastStream;
+use crate::operators::filter::StreamingFilterExec;
+use crate::operators::projection::StreamingProjectionExec;
 use crate::operators::rebatch::RebatchConfig;
+use crate::operators::unnest::StreamingUnnestExec;
 use crate::session::{DEFAULT_CATALOG_NAME, SessionManager, get_streamling_config};
 use crate::utils::batch_accumulator::AsyncBatchAccumulator;
 use arrow_schema::SchemaRef;
 use async_trait::async_trait;
+use datafusion::common::tree_node::{Transformed, TransformedResult, TreeNode, TreeNodeRecursion};
 use datafusion::common::{DFSchemaRef, Statistics, internal_err};
 use datafusion::error::Result;
 use datafusion::execution::{SendableRecordBatchStream, SessionState, TaskContext};
@@ -18,7 +22,10 @@ use datafusion::logical_expr::{
     Expr, LogicalPlan, UserDefinedLogicalNode, UserDefinedLogicalNodeCore,
 };
 use datafusion::physical_expr::{Distribution, EquivalenceProperties, Partitioning};
+use datafusion::physical_plan::filter::FilterExec;
+use datafusion::physical_plan::projection::ProjectionExec;
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
+use datafusion::physical_plan::unnest::UnnestExec;
 use datafusion::physical_plan::{
     DisplayAs, DisplayFormatType, ExecutionPlan, PlanProperties, execute_input_stream,
     execution_plan::{Boundedness, EmissionType},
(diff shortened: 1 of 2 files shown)


xiaom (approved):
LGTM!

PR #73, opened 2026-07-27
fix: flush delta meter provider on shutdown

kashive:
## Summary

Bug fix for the delta reporter flush: flush and shut down the delta meter provider on pipeline shutdown so short-lived jobs export pending delta metrics.

## Test plan

- [x] Unit regression test for delta meter provider flush on shutdown
- [x] `just fix && just lint`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4871,6 +4871,8 @@ dependencies = [
  "rand 0.9.2",
  "serde_json",
  "thiserror 2.0.18",
+ "tokio",
+ "tokio-stream",
 ]
 
 [[package]]
--- a/crates/streamling-core/Cargo.toml
+++ b/crates/streamling-core/Cargo.toml
@@ -88,3 +88,4 @@ paste.workspace = true
 
 [dev-dependencies]
 rand.workspace = true
+opentelemetry_sdk = { workspace = true, features = ["testing"] }
--- a/crates/streamling-core/src/telemetry/mod.rs
+++ b/crates/streamling-core/src/telemetry/mod.rs
@@ -4,7 +4,10 @@ pub mod recorder;
 pub mod types;
 
 pub use event_time_reader::{EventTimeReadError, EventTimeReader};
-pub use provider::{get_delta_meter, init_telemetry_provider, shutdown_test_meter_providers};
+pub use provider::{
+    get_delta_meter, init_telemetry_provider, shutdown_delta_meter_provider,
+    shutdown_test_meter_providers,
+};
 pub use types::{
     PipelineMetricMetadata, TopologyNodeType, get_global_metric_tags, set_global_metric_tags,
 };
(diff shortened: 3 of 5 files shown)


kashive:
Addressed both review findings in 36ea519d:

- **CONV-001 (regression test)**: added `telemetry::provider::tests::shutdown_delta_meter_provider_flushes_pending_counts` — builds the delta provider with an `InMemoryMetricExporter` behind a `PeriodicReader` whose interval (1h) can't tick during the test, records a `streamling_output_rows_delta` count, calls `shutdown_delta_meter_provider()`, and asserts the count was exported. Fails on the pre-fix behavior (nothing exports without the shutdown flush). Full `streamling-core` suite passes (394/394). No other test initializes `DELTA_METER_PROVIDER`, so the OnceCell set is race-free in the test binary.
- **Silent error discard**: `force_flush`/`shutdown` failures in `shutdown_delta_meter_provider` now log at `error!` with the billing-loss consequence spelled out.

kashive:
Round 2 addressed in 871b5d72:

- **CI (validate-and-test)**: was a `cargo fmt --check` failure on the long `error!` line from the previous commit — reformatted, `cargo fmt --check` now clean locally.
- **Unconditional shutdown placement**: `shutdown_delta_meter_provider()` moved outside the `if let Some(provider)` block as suggested — it null-checks internally, so it now covers any future path where the delta provider is initialized independently of the cumulative one.
- **CONV-004 (e2e billing-path test)**: deferring out of this PR. The e2e harness requires a live k3s + Prometheus environment and its metric assertions only cover scraped cumulative counters (`streamling_output_rows_total`); asserting on the OTLP-pushed delta stream needs collector-side support in the harness first. The unit test pins the exact regression mechanism (pending delta counts + shutdown ⇒ export). Happy to do the harness extension as a follow-up.
- **Test-global-state nit**: leaving as-is deliberately. No test in the binary calls `init_telemetry_provider` in prod mode (verified — test-mode init sets `SHARED_TEST_DELTA_METER`, not `DELTA_METER_PROVIDER`), and the `.expect` message names the invariant so any future collision fails loudly and diagnosably rather than silently testing the wrong provider. Adding `serial_test` as a new dependency for a currently-impossible race didn't seem worth it; testing through the real global is intentional, since "the global was never flushed" was the bug.

kashive:
## QA: PASS ✅

Validated the delta reporter flush fix with a control comparison: same short-lived job payload run once on the current prod image and once on this PR's image.

| | Control (current prod) | PR image |
|---|---|---|
| Job duration | ~31s (Completed) | ~30s (Completed) |
| Cumulative entity counts | 31 | 31 |
| **Delta entity counts** | **0 — bug reproduced** | **31 — exact match** |
| Shutdown flush log | n/a | `Shutting down delta telemetry meter provider` present, no flush errors |

Only the PR image flushed pending delta metrics on shutdown.

rafadaguiar:
lgtm

PR #75, opened 2026-07-27
feat: graceful shutdown and drain for multi-source/multi-sink pipelines

rafadaguiar:
## Summary

This PR implements a comprehensive graceful shutdown and drain mechanism for Streamling pipelines, fixing issues where multi-source/multi-sink topologies would hang indefinitely on SIGTERM (until force-killed) or lose tail data in job mode. It introduces structured shutdown coordination, terminal checkpoint finalization, a topology-derived drain policy, tracked task lifecycles across all pipeline components, and a shutdown-aware retry/lifecycle surface for plugin authors.

### Graceful shutdown infrastructure
- **New `streamling_core::shutdown` module**: a single process-wide shutdown signal (`watch` channel) that every component observes for coordinated draining; the previous per-component SIGTERM handlers are deleted.
- **`ShutdownController` + `ComponentScope`**: components spawn helper tasks through named scopes; teardown drains them in registration order across two stages (`DataPath`, `PostPlugin`), each bounded by a slice of the remaining budget. A scope that overruns its slice is named in the logs instead of hanging the process.
- **Watchdog**: a plain thread hard-exits at the edge of `STREAMLING__SHUTDOWN_BUDGET_SECS` (default 25s), so even a wedged FFI call cannot turn shutdown into a hang.
- **Drain ladder**: phased teardown — terminal-finalize wait → data-path drain → plugin cleanup → plugin terminate + dispatcher drain (bounded by `STREAMLING__PLUGIN_DRAIN_BUDGET_SECS`, default 60s) → post-plugin drain → coordinator stop.

### Drain policy (`STREAMLING__DRAIN_POLICY`: `auto` | `drain` | `fast`)
- The duplicate-vs-latency trade-off splits cleanly by topology: bounded work (job mode, hybrid backfills, bounded table/file scans) never restarts after completing, so its tail has no replay to recover it — it must drain. Plain streaming restarts by definition, so at-least-once replay covers its tail and a fast exit is safe.
- **`auto`** (default) derives the policy from the topology: drain when the pipeline has bounded work, fast-exit for plain streaming.
- **`drain`** always mints a terminal checkpoint so the drained tail's offsets commit before exit and a restart does not replay it.
- **`fast`** skips the terminal checkpoint (offsets stay uncommitted; the tail replays on restart) and caps the plugin flush wait. Sinks still flush what they consumed, scopes still drain, and the consumer still unsubscribes — a clean exit 0, not a kill. Ignored with a warning when the pipeline has bounded work.

### Checkpoint coordination
- **`CheckpointControl` handle**: bounded sources and sink futures signal lifecycle events (completion, shutdown) to the checkpoint coordinator without blocking.
- **Terminal checkpoint**: when a bounded source completes or shutdown is requested under a drain policy, a terminal epoch is minted and finalized only after all sinks ack it — no tail loss in job mode, and no replay of the drained tail on restart.
- **In-flight epoch handling**: timer epochs already in flight when the terminal checkpoint begins are retained and still finalize off late acks, while epochs that time out are swept so a lost marker cannot wedge the producer.
- **Dynamic sink tracking**: the coordinator's expected-sink set is mutable; a sink that finishes its source is dropped from the ack set so in-flight epochs finalize on the remaining live sinks instead of stalling.
- **Finalizer contract documentation**: finalizers are idempotent, may be skipped, and must never gate progress on a specific epoch.

### Shared retry crate (`streamling-retry`)
- One cancellable retry policy — bounded backoff, per-attempt timeouts, and a cancellation point between attempts — replaces the engine's assorted retry helpers, whose unbounded variants are retired.
- The same crate backs the plugin SDK's retry surface, so host and plugins give up the same way when shutdown is requested.

### Plugin SDK: shutdown signal and lifecycle (SDK 0.2.2)
- **Out-of-band shutdown signal**: the host hands every plugin a `ShutdownSignal` at load time via a new `set_shutdown_signal` module entry (generated by `init_plugin!`; the low-level example shows the manual form). SDK `shutdown::*` helpers and the SDK retry loops observe it, so a plugin's retries and waits end when the host drains — no polling, no host-side timeout games.
- **Send facade**: every host→plugin channel send goes through a shutdown-aware bounded facade — pre-shutdown a full channel is honored as backpressure, post-shutdown it is abandoned after a bounded grace with a typed error; a disconnected channel errors instead of panicking. Blocking channel sends and raw `tokio::spawn` are denied by clippy so new code stays on these paths.
- **Bounded termination**: `terminate_all_plugins` slices its send budget across plugins so N wedged input channels cannot serialize to N×5s; plugin metrics/ack/side-output forwarders drain in the `PostPlugin` stage, after plugins have flushed their final batches.
- **Preprocessor/panic hardening**: plugin preprocessor startup waits are bounded, and the panic hook never blocks.
- **Record-limit stop path**: sinks hitting `num_records_before_stop` now call `request_shutdown()` — the same graceful path SIGTERM takes — replacing the previous `SourceComplete` broadcast transport (and its replay buffer and polling probe), so tests exercise the exact drain production uses.
- **Author documentation**: a shutdown & durability lifecycle guide for plugin authors.

### Shutdown attribution diagnostics
- Plugins run on their own per-library runtime, so a wedged plugin cannot starve the host — but the host cannot see inside it either. SDK dispatchers now emit lightweight liveness markers on the metrics channel (a throttled heartbeat, plus enter/exit breadcrumbs around the checkpoint-marker flush); the host intercepts them into `std`-sync-only maps that stay readable from the watchdog's plain OS thread even if every tokio worker is stalled, so a budget blow names the wedged plugin and hook.
- A host-runtime canary (task touches a timestamp, an OS thread flags staleness) turns "is the host runtime itself starved?" into a measurement.

### Kafka connector
- **`SafeKafkaProducers`**: defers rdkafka producer teardown to detached OS threads (mirrors `SafeKafkaConsumer`), preventing deadlocks during runtime shutdown.
- **Shutdown-aware retries**: sink retry loops, startup waits, and schema-registry retries are cancellable between attempts.
- **Queue-full backoff window**: after shutdown the sink retries `QueueFull` for a bounded 10s window, letting healthy brokers finish while a wedged broker cannot pin the drain.

### Hybrid source
- Observes the shutdown signal, drains the current phase, emits the terminal checkpoint, and ends its stream instead of advancing to the next phase; helper tasks (marker forwarders, watchers) run on the component scope.

### Broadcast and multi-sink
- `BroadcastStream` retries full consumer channels for a bounded window after shutdown, then drops stalled consumers so one slow sink cannot pin every consumer of a shared scan; multi-sink nodes carry a scope so their tasks are tracked by the drain ladder.

### Other connectors
- ClickHouse source/sink, Postgres sink, dynamic tables, and external-handler operators use cancellable retry loops with per-attempt timeouts; the ClickHouse HTTP client gets connect/request timeouts; OTLP flush and WASM plugin calls are bounded.

### Lint ratchet
- The detached-spawn clippy deny covers `tokio::spawn`, `Handle::spawn`, and `JoinSet::spawn` across the connector crates and `streamling-core`; remaining allows are individually annotated as structured sites or tracked debt.

## Testing
- **E2E drain suite** (`shutdown_drain.rs`): job-mode multi-source/multi-sink pipelines drain every record into every sink before exit; streaming-mode SIGTERM under the drain policy produces a clean exit with no tail loss and no replay on restart; under the fast default, streaming SIGTERM skips the terminal checkpoint and exits promptly.
- **SIGTERM chaos suite**: fires the signal at randomized points (mid-batch, during checkpoint, during phase transitions) and asserts bounded exit, complete sink contents, no drain-budget-slice overruns, and no watchdog fire on a clean drain — including a topology with a deliberately wedged plugin, proving the shutdown signal reaches a blocked marker flush and the drain stays bounded.
- Unit coverage for the controller/scope drain ordering, retry cancellation and policy derivation, terminal-epoch finalization (retained and swept epochs), the plugin send facade, and the diagnostics maps.

## Configuration
- `STREAMLING__SHUTDOWN_BUDGET_SECS` (default 25): total shutdown budget; the watchdog hard-exits at its edge. Operators should set it below the pod's termination grace period.
- `STREAMLING__PLUGIN_DRAIN_BUDGET_SECS` (default 60): bounds the plugin terminate + dispatcher-drain phase, capped by the watchdog's remaining time.
- `STREAMLING__DRAIN_POLICY` (default `auto`): see the drain-policy section above.

## Notes
- The branch merges `main` (through the 0.3.1 bump) twice; main-side features arriving via those merges are not part of this PR's diff.
- The plugin SDK version moves to 0.2.2 for the `set_shutdown_signal` module-shape addition; both plugin examples are updated to it.

---

&gt; [!NOTE]
&gt; **High Risk**
&gt; Changes touch checkpoint finalization, Kafka offset commits on shutdown, and infinite retry loops across critical sink paths—incorrect behavior could lose tail data, duplicate on restart, or exit before drains complete.
&gt;
&gt; <sup>An earlier automated review note (Cursor Bugbot, commit 3c961c1) preceded the drain-policy, diagnostics, and SDK-signal commits; see the review thread for current findings.</sup>


rafadaguiar:
CI status note on `d42a318`: the red `claude-review` check is the review action hitting its 15-turn cap on this PR's ~12k-line diff — it errored out without posting findings, and it has concluded "failure" on every head of this PR (it also encodes HIGH/MEDIUM verdicts as job failure). All findings from its last completed review (22:14 UTC) are addressed on this head: `WatchSignal::with_deadline` now stores the `Instant` so the budget actually decays (with a regression test), the exit-path `terminate_all_plugins` no longer unwraps (a teardown error is logged and fails the exit code when the run was otherwise clean), and the kafka offset comparison uses `is_some_and`.

The substantive workflows (`streamling` incl. the full e2e suite, and `plugin-examples`) were green on `270777e` and are re-running on `d42a318`, which only adds the reviewer fixes above plus fmt.

---
_Generated by [Claude Code](https://claude.ai/code)_

rafadaguiar:
@xiaom I rolled 50% of the fleet to this image and it worked well for the last few days.
We should be good.

crates/streamling-retry/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "streamling-retry"
```
xiaom, line 2: nit: why create a new crate instead of put it into streamling-common?

rafadaguiar: Two reasons it's a separate leaf rather than part of streamling-common:

1. **It sits below common in the graph.** streamling-common itself uses these retry helpers (its Cargo.toml depends on streamling-retry), so the policy has to live at or below common anyway — a leaf with only `tokio` + `tracing` was the natural spot.
2. **It's the one crate the engine and the plugin SDK are guaranteed to share with a minimal footprint.** The SDK is compiled into every plugin dylib, and the long-term direction is to slim its dependencies (today it still pulls common, but retry-as-a-leaf means dropping common from the SDK later doesn't force re-homing the retry policy). The crate boundary also enforces the contract structurally — nobody can casually add an arrow-typed helper to the shared retry API without it showing up as a new dependency in this Cargo.toml (there's a comment in the file pinning that intent).

Folding it into common would work dependency-wise — it's a mechanical move if you feel strongly — but it would trade away the enforced-minimal shared surface.

---
_Generated by [Claude Code](https://claude.ai/code)_

xiaom (approved):

PR #77, opened 2026-07-27
feat: add array_struct_field UDF

jeffling:
## Summary

Adds a generic `array_struct_field(list, 'field') -> List<Utf8>` UDF to `streamling-common`. Projects one `Utf8` field out of every struct element of a `List<Struct>`, producing a `List<Utf8>` with the same shape.

This is the helper that was missing when reviewing #71 — there's no generic list-of-struct → list-of-strings projection anywhere in `streamling-common` or `streamling-goldsky-plugins` (the closest, `_gs_arweave_string_array`, is Arweave-specific: it concatenates two fields with a `#` delimiter and URL-encodes). With this, a nested column can feed `dynamic_table_check`'s `text[]` any-match overload without row-multiplying `unnest`:

```sql
SELECT id FROM src
WHERE dynamic_table_check('tracked', array_struct_field(items, 'account'))
```

### Design

- **Zero-copy**: output is `ListArray::try_new(field, list.offsets().clone(), values, list.nulls().cloned())` — the input's offsets buffer and row nulls are reused; only the values array is swapped for the struct's named child column.
- **Null struct elements**: a null struct has unspecified values in its child arrays, so the struct's own null mask is merged via `NullBuffer::union(structs.nulls(), child_strings.nulls())` and the string array rebuilt with `StringArray::new(child.offsets().clone(), child.values().clone(), combined)` (slice-safe; avoids `to_data`/offset pitfalls).
- **Signature**: `Any(2)`, `Immutable`. The field name must be a non-null `Utf8` scalar literal; the named field must exist and be `Utf8`. Both are user errors at execution time (the name is a runtime value, so it cannot be checked during type resolution). Missing-field errors list the available field names for debuggability.

### Test results

5/5 unit tests pass (`cargo test -p streamling-common array_struct_field`):
- field extraction preserving null list rows + empty lists + offsets shape
- null struct element → null string (the struct-validity merge)
- missing field → user error
- non-Utf8 field → user error
- empty struct-values array is safe

`just fix && just lint` clean.

### Notes

- No SQL-level test: `tokio` isn't a dev-dependency of this crate and the repo convention (e.g. `array_enumerate`) is invoke-based unit tests only.
- Only `List<Struct>` is supported (not `LargeList`); `to_large_list` exists for the reverse conversion if needed.

### DIFF
--- a/crates/streamling-common/src/functions.rs
+++ b/crates/streamling-common/src/functions.rs
@@ -2,6 +2,7 @@ use crate::functions::array_enumerate::ArrayEnumerateFunc;
 use crate::functions::array_filter::ArrayFilterFunc;
 use crate::functions::array_filter_first::ArrayFilterFirstFunc;
 use crate::functions::array_filter_in::ArrayFilterInFunc;
+use crate::functions::array_struct_field::ArrayStructFieldFunc;
 use crate::functions::byte_reverse::ReverseBytes32Func;
 use crate::functions::byte_to_hex::ByteToHexFunc;
 use crate::functions::coalesce_meta::CoalesceMetaUdf;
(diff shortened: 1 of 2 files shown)


jeffling:
Superseded — folded into #71 (commit 96fd3401 on the PR's head branch).

PR #80, opened 2026-07-28
fix: handle Arrow take offset overflow errors

jeffling:
## Summary

- route Arrow `OffsetOverflowError` results through the existing `safe_take` fallback
- preserve typed non-overflow Arrow errors
- widen the manual `List` fallback to `LargeList` so the fallback does not overflow 32-bit offsets

## Root cause

Arrow now reports some take-kernel offset overflows as typed errors instead of panics. `safe_take` only recovered from the panic path, so these errors bypassed the fallback. The manual `List` fallback also rebuilt 32-bit offsets, leaving the same overflow risk in the recovery path.

## Impact

Large offset-backed arrays and record batches can recover through the manual take path, while unrelated Arrow failures remain errors. This PR changes only the Arrow utility; batch accumulator behavior is outside its scope.

## Validation

- `cargo test -p streamling-common --lib` (248 passed)
- `just fix`
- `just lint`

### DIFF
--- a/crates/streamling-common/src/utils/arrow.rs
+++ b/crates/streamling-common/src/utils/arrow.rs
@@ -4,12 +4,13 @@
 //! with very large arrays, particularly when dealing with arrow-select's take kernel.
 
 use arrow::array::{
-    Array, ArrayRef, AsArray, LargeBinaryArray, LargeListArray, LargeStringArray, ListArray,
-    PrimitiveArray, RecordBatch, StructArray, UInt32Array, new_null_array,
+    Array, ArrayRef, AsArray, LargeBinaryArray, LargeListArray, LargeStringArray, PrimitiveArray,
+    RecordBatch, StructArray, UInt32Array, new_null_array,
 };
 use arrow::buffer::{NullBuffer, OffsetBuffer, ScalarBuffer};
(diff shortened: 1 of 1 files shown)


xiaom (approved):

xiaom:
LGTM!  

As a follow-up, we can also simplify our safe_take function by leveraging arrow's cast function. 
The proposal by codex:
 
  1. Recursively derive a large-offset target datatype:
      - Utf8 → LargeUtf8
      - Binary → LargeBinary
      - List<T> → LargeList<widen(T)>
      - Recursively widen children of LargeList, FixedSizeList, and Struct.

  2. Use `arrow::compute::cast` to convert the input array to that datatype.
  3. Retry `arrow::compute::take` on the widened array.
  4. Preserve the existing behavior of falling back only for OffsetOverflowError or the known Arrow
     list-overflow panic.

  5. Return unrelated Arrow errors unchanged.

  Conceptually:
```rust
  let widened_type = widen_offset_types(arr.data_type())?;
  let widened = arrow::compute::cast(arr.as_ref(), &widened_type)?;
  let result = arrow::compute::take(widened.as_ref(), indices, None)?;
```

PR #92, opened 2026-08-08
fix: warn and skip unregistered preprocessor ids instead of failing startup

xiaom:
Preprocessor configuration can be deployed independently from Streamling images. When shared configuration adds a new preprocessor, older images receive its id before their plugin bundle contains the implementation.

Streamling previously instantiated every configured preprocessor before evaluating the topology. An older image therefore failed at startup even when its pipeline did not need the new capability. Adding a preprocessor globally became a breaking change for every older image.

The change:

Streamling now runs only preprocessors registered by the loaded bundle. Missing ids are skipped with a warning, while available preprocessors keep their configured order. `--validate` includes the warning in its output.

If a pipeline depends on a skipped preprocessor, normal topology validation can still fail. Missing-plugin errors now list the ids provided by the loaded bundle, making an outdated image easier to distinguish from invalid configuration.

## Behavior

With an unavailable optional preprocessor in the configuration, a pipeline that does not need it now remains valid:

```json
{
  "success": true,
  "is_valid": true,
  "errors": [],
  "warnings": [
    "Preprocessor 'new_optional_preprocessor' is not registered by any loaded plugin; skipping it. Registered plugin ids: []"
  ]
}
```

## Test

Test in production with an old version plugins.


kashive (approved):
LGTM 🚀

PR #94, opened 2026-08-12
feat(clickhouse): make source page timeout and width growth tunable

jeffling:
## Problem

Hybrid sources on big tables settle into a timeout loop. A page times out, the width halves, the next page completes under the soft budget with rows far below `page_size` (default 10,000,000), the row multiplier snaps width back 4x, and the range times out again. Neither the timeout nor the growth rate could be adjusted without a rebuild.

## Changes

Two env overrides, following the Kafka source's existing `OnceCell` pattern:

| Variable | Default | Effect |
|---|---|---|
| `STREAMLING__CLICKHOUSE_SOURCE__QUERY_TIMEOUT_SEC` | `60` | Hard per-page query timeout |
| `STREAMLING__CLICKHOUSE_SOURCE__GROW_CEILING` | `4.0` | Max width multiplier per completed page |

The soft shrink budget stays derived as half the hard timeout, so lowering the timeout tightens both: pages are cancelled sooner *and* completed-but-slow pages start shrinking sooner.

Also damps the oscillation itself. The time guard in `on_complete` was gated on `elapsed > soft_time_budget`, so a page finishing at 25s of a 30s budget still grew at the full 4x ceiling. It now always caps growth at `budget/elapsed`, so width converges on the budget rather than overshooting it. A fast page is unaffected — a 1s page against a 30s budget yields a 30x multiplier, well above the ceiling — which is why the existing width-adaptation tests pass unchanged.

`page_size` and `sort_key_range` were already env-settable through the `AppConfig` `Environment` source (`STREAMLING__CLICKHOUSE_SOURCE__PAGE_SIZE`, `..._SORT_KEY_RANGE`); no change needed for those.

## Testing

- `cargo test -p streamling-connectors range_controller` — 31 passed, 0 failed
- New test `near_budget_page_growth_is_damped_by_time`: 100 rows at 20s of a 30s budget grows 1.5x, not 4x
- `cargo fmt --all --check` clean
- Neither changed file produces a clippy warning

## Note on CI lint

`cargo clippy --all-targets --all-features -- -D warnings` currently fails on 3 pre-existing `clippy::result_large_err` errors in `crates/streamling-common/src/formats/avro/schema.rs`, which this branch does not touch. Verified as pre-existing: this branch has no commits against `main` other than these two files, so `streamling-common` here is identical to `main`. Needs a separate fix (boxing `apache_avro::Error`).

### DIFF
--- a/crates/streamling-connectors/src/table_providers/clickhouse.rs
+++ b/crates/streamling-connectors/src/table_providers/clickhouse.rs
@@ -46,6 +46,7 @@ use datafusion::physical_expr::expressions::{CaseExpr, binary, col, lit};
 use datafusion::physical_plan::execution_plan::{Boundedness, EmissionType};
 use datafusion::physical_plan::projection::ProjectionExec;
 use once_cell::sync::Lazy;
+use once_cell::sync::OnceCell;
 use regex::Regex;
 use reqwest::{Error, Response};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


kashive (approved):
LGTM 🚀

PR #96, opened 2026-08-13
feat: parallel execution

sap1ens:
Streamling pipelines have run on a single stream end to end: every source exposed one partition, every sink pulled input partition 0, and per-sink "parallelism" meant slicing one accumulated batch inside `write_all`. This PR makes partitioning a first-class property of the plan — sources can emit N streams, transforms can widen or narrow them, and sinks write N streams concurrently — with checkpoint markers staying correct across every merge point.

**This PR doesn't cover every source/transform/sink though, follow-up PRs will be needed to complete support for everything** (mostly to not increase the scope even more). 

See the last section for performance impact.

## Key new operators

**`StreamingRepartitionExec`** (`operators/repartition.rs`) — a marker-aware exchange, N input partitions in, M out. Rows are placed by key hash or round-robin:

```rust
pub enum Exchange {
    Hash(Vec<PhysicalExprRef>),  // all rows of a key on one stream
    RoundRobin,                  // batches dealt out in turn
}
```

Hash placement is a correctness mechanism, not an optimization: per-key ordering and `WrappingDataSink`'s keep-last PK dedup are only correct when all rows of a key share one stream. Round-robin is therefore **never inferred** — a keyed sink missing its primary key fails planning instead of silently degrading. Internally it builds an N×M bounded-channel grid (per-(i,j) FIFO is the entire ordering guarantee; bounded channels give backpressure), with one `MarkerAligner` per output so each output emits an epoch once, not once per input. Contract: every output partition must be executed (shared readers would otherwise stall), and each exactly once (a second `execute` is an internal error, not a silently empty stream).

**`StreamingCoalesceExec`** (`operators/coalesce.rs`) — N→1, replacing stock `CoalescePartitionsExec`, which rebuilds batches with a metadata-less schema and would drop the checkpoint markers that ride on RecordBatch schema metadata.

**`ParallelSinkExec`** (`operators/parallel_sink.rs`) — replaces `DataSinkExec`, which pulls only partition 0. One `write_all` per input partition on its own task via `JoinSet`, plus the per-sink ack gate. All sink providers moved to it.

**`RepartitionNode` / `Placement`** — the logical request, resolved at planning time. `plan_repartition` elides the exchange when the input already satisfies the placement (right width + equivalent hash exprs, so a key that survived a projection or rename still matches), and lowers a single-stream target to a coalesce.

## Relies on DataFusion's existing partitioning

No parallel runtime was invented. `Partitioning` / `Distribution` / `EquivalenceProperties` / `satisfaction()` do the work; the new operators only fill the gaps DataFusion's own can't cover (schema-metadata markers, marker alignment):

- `StreamingRepartitionExec` declares `Partitioning::Hash` — the only operator in streamling allowed to, since it's the one that physically enforces placement. Elision is decided by DataFusion's `Partitioning::satisfaction()` against `Distribution::HashPartitioned`.
- Reuses `REPARTITION_RANDOM_STATE` + `create_hashes` so hashing matches stock `RepartitionExec`.
- Existing transforms (`StreamingFilterExec`, `StreamingProjectionExec`, `UnnestExec`, `CheckpointableExec`) are partition-transparent and just propagate `input.output_partitioning()`.
- Bounded file source hands `ListingTable`'s own file-group splitting straight through — the `CoalescedFileScan` wrapper is deleted.
- `EnforceSinglePartitionPhysicalOptimizerRule` — streamling replaces DataFusion's default rule set, which drops `EnforceDistribution`, so declared `Distribution::SinglePartition` requirements were never satisfied. This restores that invariant with the marker-preserving coalesce. `HashPartitioned` is deliberately unhandled (only joins/aggregations declare it, and transform validation accepts neither); keyed sink placement is planned explicitly instead. `UNION ALL` outputs stay summed and parallel.

## Sink batching

The old model — `write_batch_size = batch_size / parallelism`, then `parallel_execute` slicing one accumulated batch across N concurrent writes — is gone (`util/parallel.rs` deleted). Batching is now decided by the upstream rebatcher, and the exchange is planted **below** it, so each write stream rebatches its own share and its batches stay whole rather than being re-fragmented after the split.

- **Postgres** — per-slice writes removed; each stream writes the batch it receives. Connection pool **and** DDL behind one `OnceCell` (`IF NOT EXISTS` DDL is racy across sessions; a pool per stream would multiply connections). `records_processed` shared, so `num_records_before_stop` stays global.
- **ClickHouse** — `write_batch_size`/`parallelism` fields removed, one `send_arrow_batch` per received batch. `CREATE TABLE` behind a `OnceCell` (concurrent `CREATE TABLE IF NOT EXISTS` on replicated ClickHouse fails with "Table already exists"). `records_processed` is an `AtomicU64`. Insert/delete splitting extracted into `split_rows_by_operation`, with a zero-copy pass-through when nothing is deleted — append-only sources hit that on every batch, where the `take` was copying hundreds of MB.
- **Kafka** — N producers collapsed to one shared producer, and the hand-rolled `hash(key) % N` routing between them is deleted: `parallelism` now drives a real exchange upstream, and `ThreadedProducer` is `Sync` and batches internally. Flush and error-check happen once. `rows_received` is an `AtomicU64`.
- `deduplicate_record_batch` gained the same zero-copy fast path when no rows are duplicated.
- `AsyncBatchAccumulator::with_flush_stagger(position, total)` — `tokio::time::interval` is anchored to creation, so N streams started together flush on the same grid forever and stack their merge/copy/encode peaks into one memory spike. Each stream's timer is offset by its share of the interval.

NOTE: PG/ClickHouse changes mean the meaning of batch_size changes again - you no longer need to divide it by parallelism to get the actual batch size. **This is a breaking change.**

## Multi-sink

`MultiSinkExec` dropped its `SinglePartition` requirement (which made the optimizer coalesce a parallel source before it ever reached the fan-out). It now runs one broadcast per input partition and one `write_all` per (sink, partition), with the output being `select_all` over the per-partition consumers, and flush stagger spread across the whole N×M grid.

The sinks share one input, so they share one exchange, and it can only be keyed one way. `wrap_multi_sink_with_repartition` decides:

| group | result |
|---|---|
| all key-sensitive sinks agree on a key | exchange on it, at the widest declared `parallelism` |
| no key-sensitive sinks | round-robin, exchange only if one asked to be wider |
| sinks disagree on the key | single stream + warning |
| any sink can't be parallelized | whole group single stream + warning |

Because `MultiSinkExec` calls `write_all` directly rather than through `ParallelSinkExec`, it registers the ack gates itself. Scan sharing is also per-partition now — partition *j* of the base plan broadcasts to partition *j* of every consumer.

## Checkpointing across merge points

The invariant is that every stream carries at most one marker copy per epoch — a sink acks on first sight, and the coordinator finalizes on the first ack per sink. `MarkerAligner` enforces it wherever N streams merge (coalesce, each repartition output, a sink writing N partitions): a marker is held until every live input delivers it, and an ended input counts as delivered. Alignment only ever *delays* markers — data is never blocked, so at-least-once holds. A global `SINK_ACK_GATES` registry does the same for sinks (`register_sink_streams` / `report_marker_at_sink` / `sink_stream_done`); unregistered sinks ack immediately, i.e. exactly their old behavior. `send_checkpoint_ack` logs instead of `unwrap`-panicking, since the coordinator subscriber can exit before a sink's last stream does.

Kafka consumer instances each subscribe with their own id and unsubscribe on exit — otherwise one instance shutting down early fails the coordinator's broadcast to the rest.

## Scope and defaults

**Sources**

| source | `parallelism` | default |
|---|---|---|
| `kafka` | ✅ | **1** - all existing pipelines keep the same behavior; all instances share one consumer group so the broker assigns disjoint partitions; clamped to the topic's partition count with a warning; lag reporter runs on instance 0 only |
| `file` (bounded) | ✅ | **session `target_partitions`** — actual width is `min(parallelism, file_count)` |
| `file` (continuous) | ❌ | rejected above 1 (one watermark cursor, one checkpoint drain point) |
| `clickhouse`, `hybrid`, `plugin` | ❌ | single-stream for now |

**Transforms**

| transform | `parallelism` | notes |
|---|---|---|
| `sql` | ✅ | default = input width. Exchange goes **under** the SQL (above the upstream scan) keyed by the transform's `primary_key`, so the filter/projection run at the requested width |
| `script` | (unrelated) | its `parallelism` sizes the WASM instance pool inside one stream; the operator is `SinglePartition` |
| `handler`, `plugin`, `dynamic_table` | ❌ | single-stream for now— forces a coalesce |

**Sinks** — `parallelism` defaults to `None` = inherit the input's width, so a `parallelism: 4` source feeding a keyed sink already gets 4 concurrent write streams.

| sink | placement | `parallelism` |
|---|---|---|
| `postgres`, `postgres_aggregate`, `clickhouse`, `kafka` | `ByKey(primary_key)` | ✅ |
| `print`, `blackhole`, `memory` | `RoundRobin` (no key needed) | ✅ |
| `webhook` | `Single` | ❌ single-stream for now |
| `plugin` | `Single` | ❌ single-stream for now |

Session `target_partitions` is raised to `max(available_parallelism(), highest declared parallelism)` so the same YAML behaves the same on any machine. `parallelism: 0` and >1 on a structurally single-stream node are rejected at startup rather than silently clamped. `postgres_aggregate` parallelizes the *insert* into the landing table, not the aggregation — the aggregate table is trigger-maintained, and concurrent inserts landing in the same `group_by` bucket contend on that row.

**Still TODO** (primarily to minimize scope): parallel ClickHouse / hybrid / plugin sources, continuous file source, parallel `script`/`handler`/`plugin`/`dynamic_table` transforms, parallel webhook and plugin sinks, and multi-sink groups with conflicting primary keys (currently one stream).

## Observability

Every operator's `DisplayAs` now reports `partitions=N`, and the planned physical plan is logged at `info` before execution starts (`DataFrame::collect` split so the plan is visible even if execution fails immediately).

## Topology Examples

<table>
<tr><th>Topology</th><th>Physical plan</th></tr>

<tr>
<td>

**Keyed sink, matching widths** — sink needs to enforce keyed execution, so a hash exchange is planted at the sink edge, below the rebatcher.

<pre>
sources:
  blocks:
    type: kafka
    topic: subgraph.123456.event.account
    primary_key: id
    parallelism: 3
transforms:
  filter_blocks:
    type: sql
    sql: select * from blocks
    primary_key: vid
sinks:
  pg_even_blocks:
    from: filter_blocks
    parallelism: 3
    type: postgres
    table: even_blocks
    schema: public
</pre>
</td>
<td>
<pre>
ParallelSinkExec: partitions=3, sink=PostgresSink
  RebatchExec(batch_size=100, interval=1s, partitions=3)
    StreamingRepartitionExec: partitions=3, keys=[vid@0]
      CheckpointableExec (for KafkaSourceExec), partitions=3
        KafkaSourceExec: partitions=3
</pre>
</td>
</tr>

<tr>
<td>

**Keyless sink at the source's width** — a print sink needs no key and is already 3 wide, so the exchange is elided entirely.

<pre>
sources:
  blocks:
    type: kafka
    topic: subgraph.123456.event.account
    primary_key: id
    parallelism: 3
transforms:
  filter_blocks:
    type: sql
    sql: select * from blocks
    primary_key: vid
sinks:
  print_even_blocks:
    type: print
    from: filter_blocks
    parallelism: 3
</pre>
</td>
<td>
<pre>
ParallelSinkExec: partitions=3, sink=PrintSink (sample_every=1)
  CheckpointableExec (for KafkaSourceExec), partitions=3
    KafkaSourceExec: partitions=3
</pre>
</td>
</tr>

<tr>
<td>

**Transform widens, sink narrows** — two exchanges in opposite directions. The transform's goes *under* its SQL so the transform's own work runs at 4; the sink's is round-robin because a print sink has no ordering requirement.

<pre>
sources:
  blocks:
    type: kafka
    topic: subgraph.123456.event.account
    primary_key: id
    parallelism: 3
transforms:
  filter_blocks:
    type: sql
    sql: select * from blocks
    primary_key: vid
    parallelism: 4
sinks:
  print_even_blocks:
    type: print
    from: filter_blocks
    parallelism: 3
</pre>
</td>
<td>
<pre>
ParallelSinkExec: partitions=3, sink=PrintSink (sample_every=1)
  StreamingRepartitionExec: partitions=3, round_robin
    CheckpointableExec (for StreamingRepartitionExec), partitions=4
      StreamingRepartitionExec: partitions=4, keys=[vid@0]
        KafkaSourceExec: partitions=3
</pre>
</td>
</tr>

<tr>
<td>

**Same key at both stages** — 3→4 for the transform, 4→3 for the sink, both hashed on `id`. The uneven per-thread row counts are the three `write_all` streams running concurrently.

<pre>
sources:
  blocks:
    type: kafka
    topic: subgraph.123456.event.account
    primary_key: id
    parallelism: 3
transforms:
  filter_blocks:
    type: sql
    sql: select * from blocks
    primary_key: id
    parallelism: 4
sinks:
  print_even_blocks:
    from: filter_blocks
    parallelism: 3
    type: postgres
    table: even_blocks
    schema: public
</pre>
</td>
<td>
<pre>
ParallelSinkExec: partitions=3, sink=PostgresSink
  RebatchExec(batch_size=100, interval=1s, partitions=3)
    StreamingRepartitionExec: partitions=3, keys=[id@2]
      CheckpointableExec (for StreamingRepartitionExec), partitions=4
        StreamingRepartitionExec: partitions=4, keys=[id@2]
          KafkaSourceExec: partitions=3
</pre>
<pre>
INFO tokio-rt-worker ThreadId(10) postgres: Processing 5 rows
INFO tokio-rt-worker ThreadId(17) postgres: Processing 1 rows
INFO tokio-rt-worker ThreadId(16) postgres: Processing 8 rows
</pre>
</td>
</tr>

<tr>
<td>

**Single-stream sink** — a webhook sink is `Placement::Single`, so the parallel source is coalesced down to one write stream. This is where markers from the three branches are aligned back to one copy.

<pre>
sources:
  blocks:
    type: kafka
    topic: subgraph.123456.event.account
    primary_key: id
    parallelism: 3
transforms:
  filter_blocks:
    type: sql
    sql: select * from blocks
    primary_key: id
sinks:
  print_even_blocks:
    from: filter_blocks
    type: webhook
    url: http://127.0.0.1:8000
    one_row_per_request: true
</pre>
</td>
<td>
<pre>
ParallelSinkExec: partitions=1, sink=HttpSink (url=http://127.0.0.1:8000)
  StreamingCoalesceExec: partitions=1
    CheckpointableExec (for KafkaSourceExec), partitions=3
      KafkaSourceExec: partitions=3
</pre>
</td>
</tr>

<tr>
<td>

**Single-stream transform** — a `handler` declares `Distribution::SinglePartition`, so `EnforceSinglePartition` coalesces into it; the sink then widens back out round-robin.

<pre>
sources:
  blocks:
    type: kafka
    topic: subgraph.123456.event.account
    primary_key: id
    parallelism: 3
transforms:
  enrich_blocks:
    type: handler
    from: blocks
    url: http://127.0.0.1:8000
    one_row_per_request: true
    primary_key: id
sinks:
  print_even_blocks:
    type: print
    from: enrich_blocks
    parallelism: 3
</pre>
</td>
<td>
<pre>
ParallelSinkExec: partitions=3, sink=PrintSink (sample_every=1)
  StreamingRepartitionExec: partitions=3, round_robin
    ExternalHandlerExec: url=http://127.0.0.1:8000, partitions=1
      StreamingCoalesceExec: partitions=1
        KafkaSourceExec: partitions=3
</pre>
</td>
</tr>

</table>

## Performance Impact

Multiple partitions can execute multiple streams in parallel and better utilize CPU. Current benchmarks show that Streamling significantly underutilizes CPU when running with a single partition. 

After running some benchmarks comparing this PR with the baseline (0.3.1):

- Kafka source + Blackhole sink, 3 CPUs, parallelism of 3: **55% higher throughput**
- Kafka source + Blackhole sink, 1 CPU, parallelism of 3: **37% higher throughput**
- Kafka source + ClickHouse sink, 3 CPUs, source parallelism of 1, sink parallelism of 4: **87% higher throughput**
- File source + ClickHouse sink, 3 CPUs, global parallelism of 4: **~300% higher throughput**


jeffling:
Pushed 7e24dff3 with two changes:

- `effective_parallelism` now runs its rdkafka metadata fetch inside `tokio::task::block_in_place`, matching the other blocking rdkafka calls in this file, so planning no longer pins a Tokio worker for up to 10s.
- The shared producer scales `queue.buffering.max.messages/kbytes` by the sink's write-stream count (taken from the input plan's partition count). Collapsing N producers into one had quietly cut aggregate buffering headroom by N, meaning more QueueFull stalls under burst load. Scaling the queue restores the old capacity. Unit test included.

Two startup behavior changes from this PR that should land in the release notes:

1. `parallelism: 0` on any node now fails startup. On 0.3.x it was silently clamped to 1.
2. A postgres sink with `parallelism > 1` and no resolvable primary key now fails planning with "requires a primary key to partition by". On 0.3.x this worked because writes were sliced by row count without a key, so any keyless append-only pipeline that used `parallelism` for throughput breaks on upgrade.

One related gap for a follow-up: the primary-key check only runs when a repartition node is actually built (explicit sink parallelism or keyed placement). A keyless sink that just inherits upstream width skips it entirely. Harmless today because sources default to width 1, but if source defaults ever widen, keyless postgres/clickhouse sinks silently become N concurrent unordered writers with no warning. The check should probably fire on effective input width, not declared sink parallelism.

sap1ens:
@jeffling great changes, thank you! I hadn't realized we're tuning `queue.buffering.*` properties.

jeffling (approved):

jeffling:
This looks good to me, but since it could impact production pipelines, I think we should have an additional review

xiaom:
change LGTM. 

next steps before we merge & roll this into production:

1. run a few pipeline tests to verify 
  a. the default behaviour is the same as parallelism = 1  
  b. pipeline health check metric behaviour does not change
  c. checkpoint correctness across the various sinks
2. check that existing pipelines with parallelism > 1 / batch settings does not have any regression
3. merge and roll out slowly with our rolling upgrade tool to catch any issue

cc @rafadaguiar  @kashive

sap1ens:
Sounds like a good plan! Also, you may consider testing https://github.com/goldsky-io/streamling/pull/99 right away. We can create a stacked PR if needed.

rafadaguiar:
Left it soaking for 4 days on 10% of the fleet and it had no problems.
Will rebase, merge, and deploy to 100% today.

PR #98, opened 2026-08-19
fix: handle sliced ListArray in array_enumerate

hawkaa:
## Problem

`array_enumerate` panics on any sliced `ListArray`:

```
called `Result::unwrap()` on an `Err` value: InvalidArgumentError(
  "Incorrect array length for StructArray field \"value\", expected 3 got 5")
  at arrow-array-58.3.0/src/array/struct_array.rs:91
```

A sliced `ListArray` keeps the **full** child buffer and its offsets stay absolute into that buffer. The UDF built the `index` child from the slice's offsets (so: as many indices as the slice covers) but paired it with `list_array.values()` — the whole child — and `StructArray::from` rejects children of differing length. The result offsets were also reused verbatim, which would leave them pointing past the child once it is sliced.

Anything doing `unnest(array_enumerate(...))` can hit this; a plan that hands the UDF a sliced batch takes the pipeline down.

## Fix

Slice the child to `[first_offset, last_offset)` — the window the array actually covers — and rebase the offsets onto it with `OffsetBuffer::from_lengths`. Zero-copy child slice; the only new allocation is the offsets buffer (rows + 1), and per-row lengths are preserved so nulls and empty rows are unaffected.

## Test

`test_array_enumerate_sliced_list` builds `[["a","b"],["x","y","z"]]`, calls the UDF on `full.slice(1, 1)`, and asserts `[(0,"x"), (1,"y"), (2,"z")]`. It reproduces the panic above verbatim before the fix.

Verified: `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo test --workspace --exclude streamling-e2e` → 1131 passed, 0 failed.

### DIFF
--- a/crates/streamling-common/src/functions/array_enumerate.rs
+++ b/crates/streamling-common/src/functions/array_enumerate.rs
@@ -23,6 +23,7 @@
 //! -- 2               | tx3
 //! ```
 
+use arrow_buffer::OffsetBuffer;
 use arrow_schema::{Field as SchemaField, FieldRef};
 use datafusion::arrow::array::{Array, ArrayRef, Int64Builder, ListArray, StructArray};
 use datafusion::arrow::datatypes::{DataType, Field, Fields};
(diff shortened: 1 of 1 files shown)


rslowinski (approved):

rafadaguiar:
lgtm

PR #99, opened 2026-08-19
feat: parallel execution support for more transforms / connectors

sap1ens:
Follow-up to https://github.com/goldsky-io/streamling/pull/96. 

Adding parallel execution support for HTTP handler transform, WASM Script transform and Webhook sink.

Mostly straightforward changes following the conventions established in the main PR. 

One concern: the WASM Script transform had a default parallelism of 4. **Should we keep it?** If so, it means parallel execution will be enabled for any pipelines with the WASM Script transform.


rafadaguiar (approved):
LGTM.
Will stack on #96

PR #105, opened 2026-08-21
feat(dynamic-table): index the time column on table creation

jeffling:
## Problem

`PostgresDynamicTableBackend` creates dynamic tables like this:

```sql
CREATE TABLE IF NOT EXISTS <table> (
    "<value_column>" TEXT PRIMARY KEY,
    "<time_column>"  TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CLOCK_TIMESTAMP()
);
```

The value column gets an implicit btree from the PRIMARY KEY. **The time column gets nothing.**

That column is on the hot path whenever `cache_enabled` is set:

- `latest_update()` runs `SELECT MAX("<time_column>")` on **every** `contains()` call, including empty and all-null batches
- `load_cache()` runs `WHERE "<time_column>" > $1` on every incremental refresh

Unindexed, both are sequential scans. On a multi-million-row table that means a seq scan per record batch — the cache degrades worst on exactly the tables it exists to help. `validate_cache_time_column()` doesn't catch it either: it runs `... WHERE "<time_col>" >= CURRENT_TIMESTAMP AND FALSE`, which proves the column exists and is orderable but never scans, so enabling the cache on an unindexed table succeeds silently.

Today the README just tells you to add the index by hand.

## Change

Create the index as part of table creation:

```sql
CREATE INDEX IF NOT EXISTS "idx_<table>_<time_column>" ON <table> ("<time_column>")
```

Naming follows the existing precedent in `utils/pg.rs:445`.

Two deliberate scoping decisions:

- **Only on the creation path.** This sits inside the existing `if !table_exists` branch. Building an index on a live multi-million-row table during startup takes a `SHARE` lock and blocks writes, so tables Streamling did not create are left untouched — those still need the manual index, and the README now says so.
- **Not gated on `cache_enabled`.** The table is empty at creation so the index is free, and the flag can be switched on later.

## Verification

`cargo fmt --check`, `cargo check -p streamling-core`, `cargo check -p streamling-e2e --tests`, and `cargo clippy -p streamling-core -p streamling-e2e --all-targets -- -D warnings -A clippy::result_large_err` (matching the CI invocation) all pass.

The emitted DDL was run against `postgres:15-alpine` directly. It is valid, idempotent on a second run, and produces the intended plan on 50k rows:

```
 Result
   InitPlan 1 (returns $0)
     ->  Limit
           ->  Index Only Scan Backward using idx_indexed_members_updated_at on indexed_members
                 Index Cond: (updated_at IS NOT NULL)
```

Previously a `Seq Scan`.

`test_postgres_dynamic_table_creates_time_column_index` in `crates/streamling-e2e/tests/dynamic_table.rs` asserts the index exists on a freshly created cached table. It compiles, but I did not execute it locally — the e2e harness needs a k3s cluster and a release `streamling` binary, so the e2e job is the gate on it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -996,8 +996,9 @@ transforms:
 
 The cache is off by default and is used only when both settings are present. The initial lookup
 loads the full table through bounded PostgreSQL cursor pages. Each later `dynamic_table_check`
-batch reads `MAX(time_column)` and appends only rows newer than the cached maximum. Index the time
-column so these checks and range reads stay cheap.
+batch reads `MAX(time_column)` and appends only rows newer than the cached maximum. Streamling
+creates an index on the time column for tables it creates; add that index manually only for
+pre-existing tables Streamling did not create.
 
 PostgreSQL dynamic tables are append-only when the in-memory cache is enabled. Updating or deleting
 existing membership, including removals, is not supported in this mode. Keeping `time_column` current
--- a/crates/streamling-config/src/app_config.rs
+++ b/crates/streamling-config/src/app_config.rs
@@ -162,7 +162,11 @@ pub struct PostgresDynamicTableBackendConfig {
     pub sslmode: String,
     pub max_connections: Option<u32>,
     pub dt_schema_name: Option<String>,
-    /// Enables incremental in-memory caching for dynamic tables that explicitly set `time_column`.
+    /// Global default for incremental in-memory caching of dynamic tables.
+    ///
+    /// Individual `dynamic_table` transforms may override this per topology via
+    /// their `cache` field; when that field is omitted the global value applies.
+    /// Caching is only ever active when the table sets `time_column`.
     #[serde(default)]
     pub cache_enabled: bool,
 }
--- a/crates/streamling-core/src/dynamic_table.rs
+++ b/crates/streamling-core/src/dynamic_table.rs
@@ -68,6 +68,7 @@ impl DynamicTableBackendFactory {
         schema: Option<String>,
         column: Option<String>,
         time_column: Option<String>,
+        cache_enabled_override: Option<bool>,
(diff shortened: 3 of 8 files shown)


jeffling:
Pushed 3f1109c — the two DDL statements now run in one transaction.

Worth calling out because it wasn't just tidiness. As originally written they were two separate non-transactional `execute` calls, which leaves a hole: if `CREATE TABLE` commits and `CREATE INDEX` then fails transiently, `get_pool()` returns `Err`, but the next attempt finds `ensure_table_exists() == true` and skips the entire `!table_exists` branch. The index would never be created, and nothing would surface it — the table just stays unindexed forever, which is the exact failure this PR exists to prevent.

Postgres DDL is transactional, so both statements now either land together or not at all, and a failed init retries cleanly.

`CREATE SCHEMA` deliberately stays outside the transaction — it's separately idempotent and harmless on its own.

For the record on why the index isn't simply declared inline in `CREATE TABLE`: Postgres has no such clause. Only `PRIMARY KEY` / `UNIQUE` / `EXCLUDE` constraints create indexes inline, and none of them fit a non-unique timestamp column. (MySQL/MariaDB do support `INDEX (...)` inside `CREATE TABLE`, which is where the expectation usually comes from.) Verified against `postgres:15-alpine`:

```
ERROR:  syntax error at or near "("
LINE 1: ..."updated_at" TIMESTAMPTZ NOT NULL, INDEX ("updated_at"));
                                                    ^
```

`cargo fmt --check`, `cargo check -p streamling-core`, and `cargo clippy -p streamling-core --all-targets -- -D warnings -A clippy::result_large_err` all pass on the updated branch.

PR #106, opened 2026-08-22
perf(dynamic-table): Arrow-backed collision-free key set for the postgres cache

jeffling:
## What

Replaces the postgres dynamic table cache's `HashSet<Box<str>>` with `ArrowKeySet`, an Arrow-backed exact membership set, and adds a configurable cache freshness-check debounce so the cached path no longer pays a `SELECT MAX(time_column)` round trip on every batch.

| | before | after |
|---|---|---|
| storage | one heap allocation per key | one contiguous `LargeStringArray` |
| probe | bucket probe + pointer chase to a scattered key | single bucket probe, key inline in the buffer |
| memory (1.4M keys) | ~112 MB (estimated) | **96.8 MB (measured)** — 14% less |
| hashing | per-value SipHash-1-3 (~20-30 ns) | one vectorized ahash pass over the batch (~5 ns/key) |
| needle side | a `String` allocation per row | the Arrow array already in hand |

With 42-byte keys the string bytes dominate, so the memory win is real but modest. Measured breakdown for 1.4M keys: keys 78.3 MB + hashes 11.2 MB + table 7.3 MB = 96.8 MB, against ~112.4 MB estimated for `HashSet<Box<str>>`.

## Cache freshness debounce

Before this change, cached `contains()` ran `SELECT MAX(time_column)` on **every** invocation — one round trip per batch even when the cache was seconds old. The cached path now trusts its watermark for a configurable window:

- **`cache_refresh_debounce_ms`** — set per dynamic-table transform or globally under `dynamic_table_backend.postgres`. When both are omitted, a built-in 1000ms default applies (also to configs written before the setting existed); `0` restores re-check-on-every-batch. Documented in the README and `default_config.yaml` with the upgrade note.
- **Precedence:** topology value wins over the global value, then the built-in default; explicit `0` either way disables the window. Resolution is a pure function with direct unit tests.
- **One check per window under concurrency:** the window is claimed via a compare-and-swap, so concurrent callers on the same batch yield at most one freshness check, and a refresh is skipped entirely when the watermark is unchanged.
- **Empty and all-null batches return before touching the database.** Dataset sources emit heartbeat batches continuously; without the short-circuit the cached path measured ~117x more round trips than the uncached path in a live A/B at debounce=0.
- **Semantics:** rows written by OTHER writers may be missed for up to one window; this pipeline's own writes are mirrored into the cache immediately after commit and stay visible without waiting.
- **Logging:** a built-in-default window logs at debug; an operator-chosen non-zero window logs a `warn!` naming the staleness tradeoff.
- **Time-column index:** the index is ensured when the table is set up so `SELECT MAX` stays cheap when a check does run; a slow check logs a warning suggesting a missing index.
- **Resilience:** transient connection failures during lazy pool init are retried instead of failing the first batch.

## Cache mirror

`append()` mirrors every committed key into the cached set so this pipeline's own writes are visible immediately, without waiting for the next freshness check. The mirror (and refresh deltas) probe before extending the set: an all-duplicates append copies nothing, only genuinely-new keys reach the backing buffer, and rows already buffered — including re-fetched ones — add no bytes. `len()` keeps `HashSet` semantics (distinct count via insert-dedup), so `total_entries` / `added_entries` log fields mean what they did before.

## Measured

1.4M EVM-address-shaped keys, batches of 50, best of 3, `--release` — ns per probe:

| scenario | `HashSet<Box<str>>` | this change | |
|---|---|---|---|
| all_hit | 311 | **107** | 2.90x faster |
| all_miss | 61 | **44** | 1.41x faster |
| mixed 50/50 | 190 | **85** | 2.25x faster |
| dup_heavy (5 distinct of 50) | 64 | **43** | 1.49x faster |

The benchmark ships as an `#[ignore]`d test so this is re-checkable rather than re-argued:

```
cargo test --release -p streamling-core --lib key_set::bench -- --ignored --nocapture
```

It asserts every implementation returns identical checksums, so it is a differential correctness test as well as a timer. Run-to-run variance from binary layout alone is ~20%, so treat sub-15% differences as noise.

Dictionary-encoded columns (parquet/avro sources) are accepted: the probe hashes the K distinct dictionary values and maps answers onto the N rows with `take` — wins when K ≪ N, loses when K approaches N; the code path documents the crossover.

## Why it can't return a wrong answer

The hash **only selects candidate buckets**. Equality is decided by comparing `&str` bytes against the stored haystack key:

```rust
self.table.find(hash, |&idx| self.keys.value(idx as usize) == needle)
```

A collision costs one failed compare and probing continues. This is the same construction as DataFusion's `ArrayStaticFilter` — which is `pub(super)`, and whose public `in_list()` takes one `PhysicalExpr` per element, so it can't be reused for a set this size. arrow-rs has no membership kernel at all (arrow-ord is cmp/sort/rank/partition, arrow-select is filter/take/concat/interleave/merge), so the structure has to be ours.

Storing only hashes was rejected: 64-bit over 100M keys is a ~0.03% chance of at least one colliding pair, and a false positive here silently drops a real row.

## Correctness details worth reviewing

- **`Utf8` needle vs `LargeUtf8` haystack hash identically.** `hash_array` is generic over `ArrayAccessor` with `Item = &str` for both `StringArray` and `LargeStringArray`. It's the load-bearing invariant — if it ever breaks the cache goes silently wrong — so `utf8_and_large_utf8_hash_identically` tests it directly. The `Utf8View` divergence (inline u128 hashing hashes differently for the same content) is guarded by `utf8view_hashes_differ_from_utf8`.
- **`LargeStringArray`, not `StringArray`.** i32 offsets would cap the cache at 2 GB of key bytes (~48M keys), a regression against the unbounded `HashSet`.
- **Null slots never read a garbage hash.** `hash_array` only writes valid indices, so nulls are skipped before touching `hashes[i]`, and null needles never enter the buffer.
- **The buffer never accumulates re-offered bytes.** Extends probe first; `extend_from_reoffered_keys_do_not_grow_buffer` pins this (100 repeated appends of the same keys leave the buffer row count unchanged).
- **u32 index overflow returns an error** rather than wrapping — a wrapped index is a wrong membership answer.

## Known ceiling

`extend_from` uses `arrow::compute::concat`, so an extend with genuinely-new keys copies the whole key buffer: O(total bytes) per extend. That's the right trade for the read-mostly workload this set exists for, and wrong if appends become frequent. Flagged inline with the upgrade path (chunk vector + packed `(chunk, row)` index) rather than left silent.

## Verification

```
cargo fmt --all --check                                                    # clean
cargo clippy --all-targets --all-features -- -D warnings --A clippy::result_large_err  # clean
cargo test -p streamling-core --lib                  # 545 passed, 0 failed, 1 ignored (the bench above)
```

Unit tests cover: exact hit/miss, null preservation, the Utf8/LargeUtf8 hash invariant, the Utf8View divergence guard, incremental extend, duplicate handling, haystack nulls, a 100k-key sweep asserting every present key hits and every absent key misses (exercising real bucket collisions), debounce precedence (topology > global > built-in default, explicit-zero passthrough), the empty/all-null early return, and repeated-append buffer stability. The dynamic-table e2e pins `cache_refresh_debounce_ms: 0` explicitly, since it asserts external-writer visibility on the next batch; CI runs the full e2e suite against the branch image.


jeffling:
Addressed the latest AI review findings in e3f434e9:

**1. DefaultHasher instability (fixed)** — `build_time_column_index_name` now derives the suffix with FNV-1a (fixed algorithm, no cross-toolchain drift), and the unit test pins the exact expected name (`idx_…_54465ea85ad5121b`) so any change to the hashing algorithm fails CI rather than silently creating duplicate indexes after a Rust upgrade.

**2. Debounce check-store race (fixed)** — the load-then-store on `last_freshness_check_ms` is now a `compare_exchange` claim extracted into `try_claim_freshness_window`: concurrent callers on the same batch yield at most one `SELECT MAX` per window. The unpopulated-cache path still always runs the first load (a failed claim only short-circuits when the cache is already populated, so the `expect("cache is loaded after refresh_cache")` invariant is preserved). Covered by `freshness_window_claim_yields_at_most_one_check_per_window`.

**3. append_batch "silent" failure (no change needed — doc strengthened)** — `append_batch` wraps the insert in `retry_forever_with_backoff_async_returning`: it returns only after the batch commits (or the task is torn down). Errors are not swallowed; there is no failure path that reaches the cache mirror with uncommitted keys. The invariant the inline comment relied on is now stated explicitly in `append_batch`'s doc and referenced at the mirror site.

**4. as_any() nit (declined)** — the downcast escape hatch on the trait is the conventional pattern (used by DataFusion, anyhow, etc.) and is exercised by test mocks in `functions/dynamic_table.rs` as well as `dynamic_table.rs`; a `#[cfg(test)]` accessor would have to be replicated on every mock impl for no behavioral gain.

**5. Per-batch debug! (fixed)** — dictionary fast-path decision, plain-path backend invocation, and list-path invocation logs downgraded to `trace!` (they fire per batch in production).

Full suite green (466 lib tests), fmt/clippy clean, CI passed on e3f434e9.

kashive:
**The own-write cache mirror in `append()` grows the key buffer without bound and pays an O(total-bytes) copy per batch**

`append()` mirrors **every non-null value in the batch** into the `ArrowKeySet` after commit (`postgres.rs` — `appended_keys` is built from all of `values`, then `extend_from` runs under the write lock). Two properties of `extend_from` that are fine on the rare-refresh path become problems on this per-batch path:

1. **Duplicate bytes accumulate forever.** `extend_from` unconditionally `concat`s all incoming bytes into the backing `LargeStringArray`; the find-before-insert dedup only guards the hash table, not the buffer. Verified with a quick test: 100 duplicate appends of a 2-key set leaves `len() == 2` but 202 rows in the array, buffer 1,048 → 2,432 bytes. Since `append()` runs once per record batch via `DynamicTableSideOutput`, and the SQL insert relies on `ON CONFLICT DO NOTHING` to dedup, any pipeline whose side output re-emits existing keys leaks memory proportional to *total rows appended*, not distinct keys. In a long-running process that's an eventual OOM.

2. **Every append memcpys the whole existing buffer.** `concat` is O(total bytes), so at the PR's own reference scale (1.4M keys ≈ 78 MB) each batch containing any values pays a ~78 MB copy plus a transient 2x allocation. The `ponytail:` comment names this ceiling for refreshes, but the mirror makes it a per-batch cost.

Related: because the mirror deliberately doesn't advance `updated_at`, the next real refresh re-fetches own-written rows and appends their bytes a second time — bounded (~2x per own key), but compounding the same growth.

**Recommendation:** probe before mirroring —

```rust
// in append(), before extend_from:
let hits = cache.values.contains_array(&needles)?;
// extend only with values where hits is false
```

This is a few lines: one in-memory probe of the batch (the same probe `contains()` already does cheaply), then `extend_from` only the misses. It eliminates the unbounded growth entirely and makes the common all-duplicates append copy nothing. The O(total-bytes) copy for genuinely-new keys remains and is acceptable for now — the chunked-storage upgrade path already named in the comment covers it if append rates ever make it material. Worth one regression test asserting repeated appends of the same keys don't grow `keys.len()`.

jeffling:
Valid on both counts — confirmed and fixed (local, lands in the next push).

**Duplicate-byte leak (1):** reproduced exactly — 100 re-extends of a 2-key set left `len() == 2` with 202 rows in the buffer.

**Fix — probe before extending, inside `extend_from` itself rather than only at the `append()` call site:** the same guarantee then covers the other mutation path, `refresh_cache` → `PostgresDynamicTableCache::append` (delta re-fetches can already be buffered). `extend_from` now probes the incoming batch against the set, returns early when nothing is new (an all-duplicates append copies nothing), and otherwise concats only the misses. Null needles never reach the buffer either — they were pure waste before. The hash-table insert-dedup stays as-is for intra-batch duplicates from the initial `from_keys` build (bounded by batch size, pre-existing behavior).

Remaining per-batch cost: genuinely-new keys still pay the O(total-bytes) `concat`, plus one hash pass and one copy of the misses — O(new bytes), not O(total). That's the accepted ceiling; the `ponytail:` chunked-storage note still names the upgrade path.

Side note: the "re-fetching our own rows is idempotent" comment in `append()` was quietly wrong — the table deduped, the buffer didn't. It's now true, and reworded.

**Regression test:** `extend_from_reoffered_keys_do_not_grow_buffer` — 100 re-extends of a 2-key set assert the buffer row count never grows; a mixed extend grows the buffer by exactly the genuinely-new keys; null-only extends are no-ops; membership stays exact throughout. Fails on the old code (buffer 2 → 202), passes now. `cargo test -p streamling-core --lib dynamic_table` → 72 passed, 0 failed; `just fix && just lint` clean.

kashive (approved):
LGTM 🚀

PR #107, opened 2026-08-24
feat(kafka): support starting a source from a timestamp

sfsf9797:
## Why

`starting_offsets` accepted only `earliest` and `latest`. Recovering a pipeline from a known point in time — after a bad deploy, a schema incident, a backfill boundary — meant replaying the entire topic or skipping the backlog outright. Neither is what you want when you know the instant you care about.

## What

`starting_offsets` now also accepts an instant:

```yaml
sources:
  raw_events:
    type: kafka
    topic: app.events
    starting_offsets: 2025-08-01T12:00:00Z   # or "1754049600000"
```

| Value | Behaviour |
| --- | --- |
| `earliest` (default) | Oldest retained message |
| `latest` | End of the partition |
| A datetime | First message timestamped at or after that instant. `2025-08-01 12:00:00` and `2025-08-01` also work; no offset means UTC |
| Epoch milliseconds | Same. Must be milliseconds |

Each partition is resolved through `offsets_for_times` before consuming starts. A partition holding no message that late starts at its end, so it picks up whatever arrives next.

The same value works as `start_at` on a hybrid source's Kafka phase.

## Notable decisions

- **Parsed once, up front.** The string becomes a `KafkaStartPosition` enum in `KafkaSourceTableProvider::new` — filling the old `// TODO: use enum` — so a bad value fails config validation (and `--validate`) instead of surfacing as a librdkafka config error at execute time.
- **A seconds-sized number is rejected**, with an error naming milliseconds. Accepting `1754049600` would resolve to 1970 and rewind the source to the start of the topic — the exact opposite of "start here", and silent.
- **`auto.offset.reset` is derived from the parsed position.** A timestamp maps to `earliest`, so a partition handed over by a rebalance after startup replays rather than skips. The timestamp itself is only resolved at startup — same scope as the existing `latest` behaviour.
- **`Offset::Invalid` from the lookup is treated as "nothing that late"** (start at the end) rather than failing the source; a genuine broker error on an element still fails.

## Compatibility

- Persisted offsets keep precedence, unchanged: a partition with state in the backend always resumes from it and never consults `starting_offsets`.
- librdkafka's `smallest`/`beginning`/`largest`/`end` aliases still parse.
- One behaviour change: `starting_offsets: error` no longer loads. It previously passed validation and then silently seeked to the beginning of the topic.

## Testing

- 6 new unit tests on the parser (boundaries and aliases, epoch millis, the three datetime spellings resolving to one instant, seconds rejection, unparseable values, reset-policy fallback). All 31 kafka unit tests pass.
- `cargo clippy --all-targets` clean on connectors + core; fmt clean.
- **`kafka_json_source_starts_from_timestamp` (e2e) is written but was not run locally** — the k3s e2e stack wasn't up on my machine. It produces two batches around a cutoff and asserts only the post-cutoff rows are read. The broker-side `offsets_for_times` path is therefore unverified until this runs; worth a look before merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,29 @@ sources:
     topic: app.events
 ```
 
+**Starting position.** `starting_offsets` controls where a partition starts when the state backend holds no offset for
+it. Partitions that do have a persisted offset always resume from it, so this only applies on a first run (or to a
+partition added later).
+
+```yaml
+sources:
+  raw_events:
+    type: kafka
+    topic: app.events
+    starting_offsets: 2025-08-01T12:00:00Z
+```
+
+| Value | Behaviour |
+| --- | --- |
+| `earliest` (default) | Start at the oldest retained message |
+| `latest` | Start at the end of the partition — only messages produced from now on |
+| A datetime, e.g. `2025-08-01T12:00:00Z` | Start at the first message timestamped at or after that instant. `2025-08-01 12:00:00` and `2025-08-01` also work; a value without an offset is read as UTC |
+| Epoch milliseconds, e.g. `1754049600000` | Same as above. Must be milliseconds — a seconds-sized value is rejected rather than silently rewinding to 1970 |
+
+With a timestamp, the source asks the brokers for each partition's first offset at or after that instant before it
+starts consuming. A partition with no message that late starts at its end, so it picks up whatever arrives next.
+Partitions assigned later by a rebalance fall back to `earliest` (the timestamp is only resolved at startup).
+
 **JSON format.** Set `data_format: json` (the default is `avro`) and declare the input schema as a `column → type` map.
 Each Kafka message payload is treated as a single UTF-8 JSON object (one row); Schema Registry is not used.
 
@@ -1243,7 +1266,7 @@ The state backend plays a crucial role in Kafka source reliability:
   - On startup, Kafka source:
     - Queries state backend for stored offsets
     - Seeks to stored positions if found
-    - Starts from configured position (earliest/latest) for new partitions
+    - Starts from the configured `starting_offsets` position (earliest/latest/timestamp) for new partitions
 
 ## Error Handling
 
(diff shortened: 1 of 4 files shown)


sfsf9797:
Thanks — took the first one, pushing back on the second.

**CONN-001 (retriable): fixed** in 02d17b77. Both the `offsets_for_times` call and the per-partition error codes it returns are now `streamling_retriable_err!`. Agreed on the reasoning: it's a ListOffsets round-trip, and a malformed `starting_offsets` can't reach this path — that's rejected at source construction — so everything arriving here is broker-side and transient. This matches how `wait_for_assignment` already classifies its timeout.

One clarification on the framing: nothing in-process retries on this flag today. `main.rs:252` reads it for log/telemetry classification and the process exits `FAILURE` either way, leaving the restart to Kubernetes. So the fix improves how a blip is *classified*, not whether it gets a second chance. If we want real in-place retry here, `retry_if_retriable` would be the tool, but that'd be a broader change than this PR — the neighbouring startup calls (`assignment`, `seek_partitions`) don't retry either.

**Hybrid path: I don't think this one holds.** `HybridUnboundedSource::start_at` is passed straight into `KafkaSourceTableProvider::new` at `hybrid.rs:271`, and that constructor is exactly where the parse was added:

```rust
let start_at = match start_at.as_deref() {
    Some(value) => value.parse::<KafkaStartPosition>()?,
    None => KafkaStartPosition::default(),
};
```

The hybrid path has no separate consumer — `create_consumer` has one definition and two call sites, both in `KafkaSourceExec::execute`, reached only via the provider's already-parsed field. So a timestamp on a hybrid source is parsed and resolved identically, and the doc comment isn't overpromising. Keeping `start_at` as `Option<String>` in `topology.rs` is deliberate: it's the serde-deserialized config surface, and parsing happens one layer down where the error can be reported against a named source.

Still outstanding from the description: the e2e test hasn't been run locally, so the broker-side lookup is unverified.

jeffling:
one concern is in practice, highly compacted topics fail when searching on the message timestamp. I'm not opposed to merging this but in the past, we've had to implement a bisecting strategy to find the 'right' data. Perhaps we can test with a few prod topics and log the results first

PR #111, opened 2026-08-28
fix: give plugin metrics their own 4096-slot channel

kashive:
## Summary
- High-rate plugins (TurboPuffer with `disable_backpressure`, Solana prism, Oasis, Tealeaf) were logging `Encountered error dispatching metrics. Error: sending on a full channel` because a `0` metrics cap fell back to the data-plane default of 50.
- Metrics now default to a 4096-slot channel. Input/output stay at `plugin.channel_capacity` (50) or the plugin's explicit buffer, so data-plane backpressure is unchanged.
- Side-output metrics channels use the same 4096 default. This is a host-only change; existing plugins pick it up without a rebuild.

## Why 4096
4096 is a burst buffer, not a measured emit rate. The number that was wrong was 50 — the same default as the data-plane channel. Metrics needed their own default, large enough that a short producer burst cannot fill the channel while the host consumer is asleep.

`process_plugin_metrics` drains with `try_recv` and sleeps 1ms only when empty. Overflow happens in that hole: the plugin `try_send`s, the host is mid-sleep (or mid-`record_plugin_metric`), and the bound fills. At 50 slots, a few writes is enough. TurboPuffer emits several metrics per write; with `disable_backpressure` a ClickHouse sweep can do more than 50 samples in that millisecond.

4096 is ~80× that old bound:

- At ~8 metrics/write it holds on the order of 500 writes before drop-on-full.
- Filling it in 1ms would take tens of thousands of writes/s. That burst does not happen.
- Residual risk is a sustained emit rate above how fast the host can record — tiny-batch sweeps, not a 1ms spike. Next step then is `spawn_blocking` + `recv()`, not a bigger number.

Memory is cheap here: 4096 `PluginMetric`s (name + value + a small tag map) is on the order of hundreds of KB to ~1MB per plugin instance, not a data-plane queue of `RecordBatch`es. 1024 would probably also have stopped the 50-slot bug; 4096 is a power-of-two with more headroom and no extra config field.

## Test plan
- [x] Unit test: `resolve_channel_caps` — unset metrics → 4096; Solana-style `output=1, metrics=0` → `(50, 1, 4096)`; explicit metrics cap still wins
- [x] `just fix && just lint`
- [ ] After deploy, Groundcover `"dispatching metrics"` on `mercury-prod` should drop toward zero on Solana / Oasis / Tealeaf / API-sink sweeps


crates/streamling-core/src/plugin/side_output.rs
```diff
@@ -101,7 +101,7 @@ pub fn register_plugin_side_outputs(
         for (source_name, schema) in source_schemas {
             // Create a per-source metrics channel
             let metrics_channel =
-                PluginMetricsChannel::new(crossbeam_channel::bounded(channel_capacity));
+                PluginMetricsChannel::new(crossbeam_channel::bounded(metrics_channel_capacity));
```
xiaom: change in `crates/streamling-core/src/plugin/side_output.rs` seems unnecessary (just a renaming)

crates/streamling-core/src/plugin.rs
```diff
@@ -309,18 +309,42 @@ fn create_channels_with_caps(input: usize, output: usize, metrics: usize) -> Plu
     }
 }
 
+/// Default capacity for the plugin→host metrics channel. Distinct from
+/// `plugin.channel_capacity` (data-plane backpressure). Metrics are
+/// best-effort `try_send`; a 50-slot bound overflows on high-rate plugins.
+pub const DEFAULT_PLUGIN_METRICS_CHANNEL_CAPACITY: usize = 4096;
```
xiaom, line 319: > a 50-slot bound overflows on high-rate plugins

can you elaborate what does this mean?

kashive: I think it just meant that the channel capacity of 50 which was the default used before overflows on plugins with high throughput

crates/streamling-core/src/plugin/side_output.rs
kashive: **Investigation** — The `side_output.rs` hunk is rename-only. The real change is the `lib.rs` call site switching from `app_config.plugin.channel_capacity` (default 50) to `DEFAULT_PLUGIN_METRICS_CHANNEL_CAPACITY` (4096). Control flow in this file was already “use the `usize` you were given.”

**Decision** — Revert the rename. The new parameter name was a guardrail so this argument is not re-tied to the data-plane knob, but it is not required for the 4096 wiring and was the only change in this file.

**Fix Summary**
- Commit: `dab97a5874be8dba6a217ba5d05a50d53796cbb9`
- Files: `crates/streamling-core/src/plugin/side_output.rs`
- Restored `channel_capacity` on `register_plugin_side_outputs`. `lib.rs` still passes 4096.

crates/streamling-core/src/plugin.rs
kashive: **Investigation** — “50-slot bound overflows” meant the old metrics fallback was `plugin.channel_capacity` (default 50). Plugins emit with non-blocking `try_send`; when that channel is full the sample is dropped and we log `Encountered error dispatching metrics` (`sending on a full channel`). A `0` metrics cap (every plugin today) used to inherit that 50.

**Decision** — Rewrite the rustdoc on `DEFAULT_PLUGIN_METRICS_CHANNEL_CAPACITY` so that explanation lives next to the constant, not only in this thread.

**Fix Summary**
- Commit: `883c3f5c8bd7329dde87431bda25b3d72e70e0d2`
- Files: `crates/streamling-core/src/plugin.rs`
- Doc now names the data-plane default of 50, `try_send` drop-on-full, and the old `0` → 50 fallback. Capacity and `resolve_channel_caps` are unchanged.

kashive:
## Babysit iteration 1

Review comments from this pass:

- **side_output rename** ([discussion](https://github.com/goldsky-io/streamling/pull/111#discussion_r3887001941)) — reverted unused `metrics_channel_capacity` rename; `lib.rs` still passes 4096. `dab97a58`
- **“50-slot bound overflows”** ([discussion](https://github.com/goldsky-io/streamling/pull/111#discussion_r3887003101)) — rustdoc on `DEFAULT_PLUGIN_METRICS_CHANNEL_CAPACITY` now names the old 50 default, `try_send` drop-on-full, and the `0` → 50 fallback. `883c3f5c`

Ignored / deferred:

- **CONV-004 e2e flood test** ([comment](https://github.com/goldsky-io/streamling/issues/111#issuecomment-5452103113)) — CONV-004 does not apply (host channel default, not a pipeline feature). CONV-001 is already met by `resolve_channel_caps_uses_metrics_default_when_unset`. A high-rate e2e would be racy and is left as a possible follow-up, not this PR.

kashive:
## Babysit complete — merge-ready

**Iterations:** 1 (plus CI wait after push)

**CI:** All required checks passed on `883c3f5c` (`validate-and-test`, `e2e-tests`, `claude-review`, release builds). Main-only publish jobs skipped, as expected on a feature branch.

**Issues addressed**

| Source | Outcome | Commit |
| --- | --- | --- |
| [side_output rename](https://github.com/goldsky-io/streamling/pull/111#discussion_r3887001941) | Reverted unused `metrics_channel_capacity` rename; `lib.rs` still passes 4096 | `dab97a58` |
| [“50-slot bound overflows”](https://github.com/goldsky-io/streamling/pull/111#discussion_r3887003101) | Rustdoc on `DEFAULT_PLUGIN_METRICS_CHANNEL_CAPACITY` now names the old 50 default, `try_send` drop-on-full, and the `0` → 50 fallback | `883c3f5c` |
| [CONV-004 e2e flood](https://github.com/goldsky-io/streamling/issues/111#issuecomment-5452103113) | **Ignored** — CONV-004 does not apply (host channel default). CONV-001 already met. Latest Claude review has no findings. | — |

**Deferred / follow-up:** none required for merge. A high-rate plugin metrics e2e remains optional later if overflow persists after deploy.

xiaom:
lgtm

PR #113, opened 2026-08-31
fix(plugin): `sleep(1ms)` instead of `yield_now()` when plugin channels are empty

kquintal:
Webhook pipelines sit at [100% regardless of resource size](https://linear.app/goldsky/issue/BDA-2962/investigate-edgeevm-webhook-pipelines-saturating-cpu). 

If increasing resources to a turbo pipeline doesn't help, must be an internal data management issue. Seeing `identical throughput AND identical p50/p95 event lag` was a big hint.

Solution: in four locations, apply a small `sleep(Duration::from_millis(1))` instead of `yield_now()`, which reschedules immediately so the runtime never gets to park. The loop re-checks its channel every 1 ms instead of spinning. 1 ms matches the two loops in the repo that already do this.

Ran some flamegraphs + benchmarks locally, surprisingly good results (see AI summary below).

# fix(plugin): park instead of spinning when plugin channels are empty

One chain, ~685 log rows/s near tip, blackhole sink, `/usr/bin/time -l`:

| plugin transform nodes | before | after |
| --- | --- | --- |
| 0 | 0.032 cores | **0.008** |
| 1 | 0.286 cores | **0.013** |
| 2 | 1.292 cores | **0.039** |

Adding a second node costs +1.0 core before and +0.026 after. That's the whole
"resizing the pipeline changes nothing" symptom: the burn scaled with node count
and wall time, not with data.

It helps under load too, which I didn't expect — the spin was stealing cycles
from real decode. Same 3000-block backfill: **14.5 → 6.5 ms CPU/block**, 161M →
93M instructions/block, 11.3M → 0.15M involuntary context switches.

## The change

A 1 ms park instead of an immediate reschedule. Not a new idiom — six loops in
the repo drain a plugin channel with `try_recv()` in an async loop, and two of
them already parked:

| loop | on empty | existing / new |
| --- | --- | --- |
| `telemetry.rs:87` — plugin metrics | park 1 ms | existing |
| `table_provider.rs:558` — checkpoint-ack forwarder | park 1 ms | existing |
| `table_provider.rs` — `PluginSourceExec` output | `yield_now()` → park 1 ms | new |
| `operator.rs` — `PluginExec` output | `yield_now()` → park 1 ms | new |
| `dispatch.rs` — `TransformPluginDispatcher` input | `yield_now()` → park 1 ms | new |
| `dispatch.rs` — `SinkPluginDispatcher` input | `yield_now()` → park 1 ms | new |

The `telemetry.rs` park dates to the open-source import (`ff0927fd`); the
`table_provider.rs` one was written three weeks ago in #61. Both patterns have
coexisted since day one and only differ under a profiler, which is probably why
this went unnoticed. (`kafka.rs:1578` also matches `Empty`, but it breaks out of
its loop rather than waiting — left alone deliberately.)

The sleep only runs when there is nothing to do; under load `try_recv` succeeds
and it is never reached. Worst case is 1 ms of extra latency per hop on an idle
pipeline.

That includes the checkpoint path in `operator.rs`, which buffers markers on the
same loop: reacting to an epoch can now take up to 1 ms longer. No message is
dropped and barrier alignment is unchanged. For scale, the ack round-trip in
these runs was 310-640 ms against a 5 s checkpoint interval, and the ack
forwarder one file over has polled at this same 1 ms since #61.

`transform_dispatcher_parks_instead_of_spinning_on_empty_input` guards it by
counting how the dispatcher waits — on the old code it records 1525 yields in
20 ms, now it records 0.

## For reviewers

- **No ABI change.** `PluginAsyncRuntime::sleep` was already on the trait, just
  unused. A patched plugin runs against an unpatched engine — I measured that
  combination at 0.33 → 0.10 cores, so the two halves can ship independently
  (the plugin side is ~71% of the win).
- **The two `streamling-core` sites have no seam for a test double**, so they
  are covered by the measurement rather than a unit test. They are the identical
  one-line change as the two in `dispatch.rs`.
- **E2E not run** — needs the k3s env, which I don't have set up locally.
  `cargo test --workspace --lib` is green (1155 tests), `just lint` clean.


kquintal:
Sink test added, constant collapsed into one `pub const` in `ffi.rs`.

Can't do the last one — those two arms call `tokio::time::sleep` directly with no runtime handle in scope, so there's nothing to count, and yield vs sleep is behaviourally identical anyway (only CPU differs). The two dispatcher tests do fail on the old code. Leaving it for a human call.

rafadaguiar (approved):

hawkaa:
@rafadaguiar could this issue be solved by some overarching "plugin architecture" work? Feels like sleeping for 1ms is not the right fix? I guess without it the pipelines are live locked?

rafadaguiar:
Most likely yes, but this looks like a quick win from where it is right now.

PR #118, opened 2026-09-03
fix(hybrid): defer phase-state persistence until a covering checkpoint finalizes

rslowinski:
## Problem

The hybrid source persisted `completed_phases: [true]` and the seeded Kafka offsets the instant the bounded stream ended (`advance_to_next_phase` → `save_state()` + `seed_offsets()`), before any sink acknowledged a checkpoint. If the sink failed inside that window (seen in production: topic ACL denied at first start, process exited 2s after the handover), every restart skipped the bounded phase and the rows it had read but the sink never wrote were lost silently. Only a full state reset recovered them.

This violates the documented rule that a source commits its position only after every sink has durably written the records before the Marker. The ClickHouse source (split on `Finalizer`) and the Kafka source (commit on `Finalizer`) already follow it; the hybrid phase pointer did not.

## Fix

- `advance_to_next_phase` changes in-memory state only.
- Kafka seed offsets are held in memory (`KafkaSourceTableProvider` / `KafkaSourceExec`). Seek order at startup: persisted offset → seeded offset → `starting_offsets`. `persist_offsets_if_absent` writes a given set put-if-absent.
- `DeferredPersistGate` (one mutex): the execute loop arms it after a phase advance with a wall-clock stamp and a **snapshot** of the state that advance produced. The existing checkpoint forwarder feeds it Markers; the first Marker created after the stamp names the earliest covering epoch, and that epoch's `Finalizer` persists the snapshot (`persist_phase_state(&snapshot)` = seeds from the snapshot + hybrid state). Writing the snapshot rather than the live state matters with several bounded phases: the next phase may already have advanced in memory without being covered. Arm generations are monotonic, so a persist finishing after a newer advance cannot disarm it.
- A finalized terminal checkpoint (job completion or shutdown) also persists an armed transition. Its Marker is pushed inline after every row and never travels on the coordinator channel, so the forwarder cannot see it.
- Spec updated (`docs/hybrid-source-state.allium`): `AdvanceFromBoundedPhase` no longer persists; new `PersistPhaseStateOnFirstCoveringFinalizer` (snapshot, terminal, retry, re-arm); `SeedIsNotDurable` guarantee.

## Behaviour change

Phase state becomes durable up to one checkpoint interval later. A restart inside that window replays the bounded phase (duplicates, at-least-once) where it previously skipped it (loss).

## Tests

- `test_advance_to_next_phase_defers_persistence`; six `DeferredPersistGate` tests (pre-stamp Markers ignored, lower epochs ignored, re-arm race, snapshot isolation, generation reuse, terminal); `seed_offsets_stay_in_memory_until_persist_seeded_offsets`.
- `fm6`, `fm1`, and both `save_state` retry tests now target `persist_phase_state`, the durable step.
- `cargo test -p streamling-connectors --lib -- table_providers::hybrid table_providers::kafka`: 84 pass. Clippy clean. `cargo check --workspace --all-targets` clean.
- e2e not run locally (needs the cluster). `test_hybrid_source_checkpoints_after_unbounded_transition` uses the `hybrid_source:%` key as its transition signal; it now appears on the first Finalizer (1s interval there) rather than immediately. `test_hybrid_clickhouse_resume_from_high_saved_split` deletes the hybrid key before run 2; the job's terminal checkpoint still writes it, so that step still exercises the fallback path.

## Known gaps

- `persist_offsets_if_absent` get→put is not atomic against the Kafka commit loop on the same Finalizer; a seed can land over a fresher commit. Replays at most one epoch of that partition; documented in the method comment.
- No unit test drives the execute loop's deferral end to end (`test_hybrid_source_execute_terminates_in_job_mode` runs with `checkpoint_control: None`); the gate is tested in isolation and the loop wiring only by the e2e hybrid suite. No e2e covers the exact incident (sink unavailable at handover, then restart). Both worth follow-ups.
- Eligibility compares the coordinator's `created_at_ms` with a stamp from the same `now_ms()`; a backwards clock step only delays persistence.

### DIFF
--- a/crates/streamling-connectors/src/table_providers/hybrid.rs
+++ b/crates/streamling-connectors/src/table_providers/hybrid.rs
@@ -1,5 +1,6 @@
 use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
@@ -831,6 +832,29 @@ impl HybridTableProvider {
             }
         }
         drop(state);
+        // Deliberately NOT persisted here. The rows the finished phase emitted
+        // are not durable until a sink-acknowledged checkpoint covers them;
+        // persisting `completed_phases` now would make a restart skip them if
+        // the sink fails first (observed: sink unavailable at first start).
+        // `persist_phase_state` runs from the execute loop on the first
+        // Finalizer whose Marker was created after this transition.
+        Ok(())
+    }
+
+    /// Durably record the current phase state plus the Kafka seed offsets
+    /// handed over at the last transition. Only call once a checkpoint that
+    /// covers everything emitted before the transition has finalized.
+    pub async fn persist_phase_state(&self) -> DataFusionResult<()> {
+        if let Some(wrapping_provider) = self
+            .config
+            .unbounded_source
+            .downcast_ref::<WrappingSourceTableProvider>()
+            && let Some(kafka_provider) = wrapping_provider
+                .get_inner()
+                .downcast_ref::<KafkaSourceTableProvider>()
+        {
+            kafka_provider.persist_seeded_offsets().await?;
+        }
         self.save_state().await
     }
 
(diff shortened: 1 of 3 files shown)


rslowinski:
Self-review pass found one gap in the first commit, fixed in `00d42bba`:

**Quiet-partition seed durability.** The Kafka Finalizer commit path skips partitions whose `consumer.position()` is `Offset::Invalid` (nothing arrived since the seek). Before this PR that was harmless because seeds were persisted eagerly for every partition. With seeds in memory only, a partition that stayed quiet until the first Finalizer had no persisted offset unless the hybrid's deferred `persist_seeded_offsets` ran; a crash between the Kafka commit and that write (or a failed write) would make the restart re-derive a *fresh* seed and skip anything delivered on that partition in between. The Finalizer path now substitutes the in-memory seed for `Invalid` partitions, so seeds become durable on the same Finalizer as the committed offsets and `persist_seeded_offsets` is a belt-and-braces no-op in the common case.

Other things I checked and consider fine, for the reviewer's benefit:
- Eligibility uses the coordinator's `created_at_ms` (same `now_ms()` clock as the hybrid stamp); a Marker created in the gap between the last bounded batch and the stamp is treated as *not* covering, which only delays persistence by one interval.
- Persist failure keeps the stamp armed and retries on the next Finalizer; a second transition before persistence re-arms with the latest state (cumulative, so nothing is lost).
- Job mode: the terminal Marker travels inline, never through the coordinator channel, so the hybrid key is not written for jobs anymore. Jobs never restart (`backoff_limit: 0`), and e2e scenario 4 deletes that key manually before run 2, so this is a no-op there.
- `test_hybrid_source_checkpoints_after_unbounded_transition` waits for the `hybrid_source:%` Kafka key as its transition signal; it now appears on the first Finalizer (interval 1s in that test) instead of immediately.

Unit tests: 77 (hybrid 35 + kafka 42) pass; clippy clean. e2e still needs the CI stack.

PR #119, opened 2026-09-04
fix(plugin): end the source stream when a plugin completes on its own

bsidelinger912:
## Problem

Job-mode pipelines whose source is a **plugin** never terminate ([FOU-1166](https://linear.app/goldsky/issue/FOU-1166/solana-job-mode-pipelines-never-terminate-checkpoint-epochs-time-out)). The fetch side completes and stops at the configured bound, but the job stays `Running` forever, cycling checkpoint epochs that time out at 300s each. **Reproduced in prod and in the local kind cluster on stock images** (bounded `solana.token_transfers`, blackhole sink, `job: true` — plus prod controls with numeric `end_block` and `end_at: latest`, all hanging identically while a bounded EVM hybrid-source job sails to `Succeeded` in ~2.5 min).

An instrumented plugin build (logging every dispatcher loop iteration, marker/finalizer delivery, and the serve pointer) showed the plugin side doing everything right:

```
04:25:34  marker epoch=1 slot_at_marker=444150601 end_block=Some(444150600) capture_end=true
04:25:34  terminal epoch CheckpointEpoch(1) finalized; setting running=false
04:25:34  source dispatcher loop exiting at iteration 307 (is_running=false)
04:25:34  source dispatcher start() returning Ok          <- clean, designed termination
04:26:04  Sending checkpoint with epoch: 2
04:31:10  Checkpoint epoch 2 timed out after 300s ...     <- host hangs anyway, forever
```

## Root cause

`PluginSourceExec`'s forwarding loop had exactly one exit: the process-wide shutdown signal. A live plugin sends an empty keep-alive `NextBatch` every ~100ms, and the loop only reaches its coordinator-marker-forwarding branch *after receiving a plugin message*. Once a bounded plugin finishes and its dispatcher exits:

1. the loop idles forever with no exit condition,
2. coordinator markers stop being forwarded to the plugin — matching the coordinator's "its Marker may never have reached a live source" and the observed disappearance of `Processing checkpoint marker` from epoch 2 onward,
3. the DataFusion stream never ends, so sinks never drain, and the run loop's terminal condition ("all sinks drained") is never met — it deliberately treats a dispatcher exiting on its own as non-terminal.

This affects **every self-terminating plugin source**, not just Solana: all `sequential_source`-based sources with `end_number` (stellar, near, oasis, hypercore, polkadot, ethereum-traces) flip `is_running()` the same way. Unbounded plugin sources are unaffected (they only stop via host `Terminate`, which uses the working shutdown-drain path).

## Fix

The host cannot distinguish "plugin idle" from "plugin done" by watching the output channel, so the plugin now says so explicitly. **The wire format is adopted from #75** (the comprehensive graceful-shutdown/drain PR, which fixes this same hang as part of a much larger change): the source dispatcher announces completion by sending **`Terminate` on its OUTPUT channel** — a plugin→host direction no plugin has ever used — and `PluginSourceExec` ends the source's record-batch stream on receipt, falling through the existing pending-marker flush (synthetic final batch to sinks) and the #61 terminal-checkpoint drain.

**Why reuse `Terminate` instead of adding a variant** (our first cut added `PluginMsg::Complete`; #75's analysis is correct and we verified it against abi_stable 0.11.3): abi_stable's nonexhaustive layout check rejects loading a library whose enum carries fewer variants than the loading side expects (`TooManyVariants`). A new variant would make a newer engine **refuse to load every plugin dylib built against the older SDK** — breaking pipelines with pinned `config_overrides.plugin_image_tag`, plugin-image rollbacks, and external plugin authors who upgrade the engine binary independently. Reusing the existing variant changes no layout: both mixed-version directions load fine. An older engine simply ignores the output-direction `Terminate` (its match falls through), retaining the old hang until upgraded — which a new variant could not have avoided either, since an old engine can't act on a message it doesn't know.

Matching #75's wire format also means whichever PR merges second rebases onto compatible semantics instead of reconciling two competing protocols.

The dispatcher announces completion on **every** exit — error exits included (best-effort, error still propagated) — so the "every dispatcher exit announces completion" contract is unconditional and a future graceful-error refactor cannot reintroduce the hang. (Today an error exit panics in the plugin worker and fails the pod loudly; job pods run `backoffLimit: 0`, so errors surface as `Failed`, not hangs.)

## Verification

**Confirmed in the local kind cluster**: with the fixed engine + plugin images loaded, the exact repro that previously hung for 40+ minutes (and 7+ hours on older attempts) reached **`1/1 Completed` in 35s**:

```
05:17:54  terminal epoch CheckpointEpoch(1) finalized; setting running=false
05:17:54  source dispatcher announcing completion to host
05:17:55  PluginSourceExec: source reported completion; ending its record-batch stream
05:17:55  All plugin dispatchers drained cleanly
```

No epoch-2 mint, no 300s timeout loop, clean process exit — first Solana job-mode pipeline to ever complete in that cluster. Also verified together with [FOU-1165](https://linear.app/goldsky/issue/FOU-1165/featstreamling-add-end-at-latest-support-for-solana-source) / streamling-goldsky-plugins#263 (`end_at: latest`): tip pinned at init, source stopped at the pinned bound, job completed — unblocking FOU-1106.

**Tests:**
- `source_dispatcher_announces_complete_on_graceful_exit` — bounded source stops itself; the completion announce is the final output message.
- `source_dispatcher_announces_complete_when_never_running` — already-stopped source still announces, and `initialize()` is not called.
- `source_dispatcher_announces_complete_on_error_exit` — an error exit still announces before the error propagates.
- `plugin_source_stream_ends_when_plugin_reports_complete` — exec-node regression test: stream ends on the announce while still delivering the batch sent before it.
- **e2e** `test_job_mode_bounded_plugin_source_completes` — a real job-mode pipeline with a bounded plugin source (`random_source` + new opt-in `stop_when_exhausted`) runs to completion; terminates on serve-completion without finalize-gating (the `sequential_source` timing), which also exercises the synthetic-final-batch flush. Full `shutdown_drain` e2e suite passes (5/5).

## Relationship to #75

#75 implements this same termination handshake (same wire format, by design — we adopted theirs) inside a comprehensive shutdown/drain framework: bounded completion-send policy, drain ladder, liveness attribution, and more. This PR is the minimal targeted fix so FOU-1106 can unblock now; since it lands the termination core in #75's own shape, #75's eventual rebase in these files keeps its additions rather than reconciling competing protocols.

## Follow-ups (not in this PR)

- Prod verification via `config_overrides` image pins on the FOU-1166 repro config.
- Verify with a bounded `sequential_source` chain (stellar/near) on real infra.
- `topology_validation.rs` still admits unbounded plugin sources to `job: true` with no diagnostic; the strict `>` in the Solana source's `should_capture_end_epoch` remains a latent hazard (tracked on the Linear ticket).

Linear-issue: [FOU-1166](https://linear.app/goldsky/issue/FOU-1166/solana-job-mode-pipelines-never-terminate-checkpoint-epochs-time-out)


bsidelinger912:
**Combined verification with the feature this was blocking:** [FOU-1165](https://linear.app/goldsky/issue/FOU-1165/featstreamling-add-end-at-latest-support-for-solana-source) / goldsky-io/streamling-goldsky-plugins#263 (`end_at: latest` — dynamically resolving and pinning the current slot as the end bound) was tested together with this fix on the local kind stack: the preprocessor pins the tip at init, the source stops at the pinned bound, announces `Complete`, and the job reaches `1/1 Completed`. This PR is the missing piece that makes that feature usable — `end_at: latest` correctly bounds the fetch, but without this fix the bounded job still never terminated (that's FOU-1166's "PR #263 does not fix this and is not the cause"). Together they unblock bounded Solana backfills for wallet feeds (FOU-1106).

bsidelinger912:
Re: the latest review — all three findings addressed in e4122c8:

**Error-path gap:** taken, with one correction to the severity claim. "The hang persists on unhappy paths" is not current behavior: an `Err` out of `start()` hits the `error!` + `panic!` wrapper in `streamling-plugin/src/lib.rs`, the panic crosses the FFI boundary as `FfiPoll::Panicked`, and the host re-panics polling the execution future — the process dies loudly, and job-mode pods run with `backoffLimit: 0` / `restartPolicy: Never`, so the Job flips to `Failed` immediately. No silent hang. The fix is still worth having as defense-in-depth: `start()` now announces `Complete` on every exit (error included, best-effort, error still propagated), so the "every dispatcher exit announces Complete" contract is unconditional and a future graceful-error refactor (the run loop already has a `fail_drain` arm for plugin futures resolving `Err`) cannot reintroduce the hang. A drop-guard can't run the async send-with-retry, so the loop was split into a `run()` helper instead — same guarantee, idiomatic shape. Covered by `source_dispatcher_announces_complete_on_error_exit`.

**ABI comment:** clarified as suggested — an older host silently discards the message and the mixed-version hang persists until the host updates.

**Dockerfile pin:** moot — `Dockerfile.local-build` is removed from the PR entirely (kept as a local dev tool; it'll come back with proper docs separately).

Also added the e2e coverage requested in the previous review (CONV-004): `test_job_mode_bounded_plugin_source_completes` runs a real job-mode pipeline with a bounded plugin source (`random_source` + new opt-in `stop_when_exhausted`) to completion — terminating on serve-completion without finalize-gating, i.e. the sequential_source timing, which also exercises the synthetic-final-batch flush. Verified locally: the new test passes in ~13s, and the full `shutdown_drain` e2e suite is 5/5.

crates/streamling-core/src/plugin/table_provider.rs
```diff
@@ -857,4 +868,63 @@ mod tests {
         unsubscribe(CHECKPOINT_COORDINATOR_CHANNEL, coordinator_sub_id);
         fake_plugin.join().expect("test plugin thread panicked");
     }
+
+    /// Regression for the job-mode hang (FOU-1166): a bounded plugin source
```
jeffling: can you remove internal references from the open-source repo?

bsidelinger912:
**Refactor in 363e202: adopted #75's wire format — `Terminate` reused in the plugin→host direction instead of the new `PluginMsg::Complete` variant.**

While comparing this fix with #75 (which solves the same bounded-source hang inside its broader shutdown/drain framework), we verified its ABI analysis against abi_stable 0.11.3 and it's correct — and it invalidated our original design:

```rust
// abi_stable-0.11.3/src/abi_stability/abi_checking.rs:682
t_exhaus.is_nonexhaustive() && t_fcount.len() > o_fcount.len()  // => TooManyVariants
```

A `WithNonExhaustive` enum fails the load-time layout check when the *loading* side knows more variants than the loaded library carries. So the `Complete` variant would have made a newer engine **refuse to load any plugin `.so` built against the older SDK** — breaking pipelines with pinned `config_overrides.plugin_image_tag`, plugin-image rollbacks during incidents, and external plugin authors upgrading the engine binary on their own schedule. (Our kind verification never caught it because we always loaded freshly built engine + plugin together; the direction we'd analyzed — old engine + new plugin — is the one that happens to load fine.)

The refactor sends `Terminate` on the source dispatcher's **output** channel instead — a direction no plugin has ever used, so it's a pure semantic addition with zero layout change: both mixed-version directions still load, and an old engine ignores the message (keeping the pre-fix hang until upgraded, which no design could avoid — the signal has to exist in the `.so` either way). Host-side handling, error-exit guarantee, and all tests are unchanged in behavior; unit + exec + e2e drain suite (5/5) pass on the retargeted code.

Credit to @rafadaguiar — #75 both identified this hazard and picked the right escape hatch; this PR now uses its exact wire format (including the log line), so whichever lands second rebases onto compatible semantics rather than reconciling two competing protocols.

bsidelinger912:
NOTE: we're going to take Raf's PR in favor of this: https://github.com/goldsky-io/streamling/pull/75, but keeping this open in case that gets held up for some reason.

PR #126, opened 2026-09-08
perf: build source providers concurrently and share HTTP clients

rafadaguiar:
## Summary

Pipeline startup, and therefore `--validate`, constructed sources one after another, and construction is where we talk to the outside world: one schema-registry fetch per Kafka phase and four ClickHouse probes per bounded hybrid phase, each a blocking round trip. A six-source union paid for ~30 sequential round trips before planning started. On top of that every `ClickHouseClient` and every `SrSettings` built its own `reqwest::Client`, and each build parses the system CA bundle (twice, via openssl): a hybrid source built five clients, a six-source pipeline 31, which was over a second of pure CPU on the validator pod.

## Changes

- **`crates/streamling/src/lib.rs`**: new `build_source_providers` step constructs the Kafka, ClickHouse, hybrid and file providers concurrently (at most `MAX_CONCURRENT_SOURCE_BUILDS = 8` at a time; the synchronous constructors run on the blocking pool since they block internally via `block_in_place`/`block_on`). The registration loop is unchanged apart from consuming the prebuilt providers, and it now iterates sources in name order. Errors are reported in name order too, so a failing topology surfaces the same error every run instead of whichever source lost the race. Plugin sources still build sequentially (fast FFI, process-wide registries).
- **`crates/streamling-connectors/src/table_providers/clickhouse.rs`**: all `ClickHouseClient`s share one process-wide `reqwest::Client`. Credentials, database and compression stay per client; the pool settings carried no per-connection state.
- **`crates/streamling-config/src/app_config.rs`**: `KafkaConfig::get_schema_registry_settings` caches `SrSettings` per (url, username, password). The lock is held across the build so concurrently constructed sources cannot all miss the cache at once.

Only the *constructors* move off the sequential path. Every builder that wires a source into shutdown and checkpointing (`with_scope`, `with_checkpoint_control`, `with_shutdown`) still runs in the registration loop, in name order, exactly where #75 put it.

## Measurements

Fake schema registry + fake ClickHouse (real Arrow IPC responses), per-call latency injected at the fakes, idle 4-core box:

| Topology | 500 ms/call before | 500 ms/call after |
|---|---|---|
| 1 hybrid dataset → ClickHouse sink | 3.2 s | 2.7 s |
| 6 hybrid datasets → union → ClickHouse sink | 17.1 s | 2.7 s |
| 6 kafka datasets → union → ClickHouse sink | 4.0 s | 0.8 s |

At 0 ms/call the six-hybrid validate drops from 1.7 s to 0.24 s (that part was the CA-bundle parsing). TLS client builds per six-hybrid validate: 31 → 2. Validation output (valid flag, warnings, error text on the failure path) is identical before and after.

## Behavior notes

- Sources are now registered in name order instead of `HashMap` order. Registration is order-independent (session catalog, primary-key registry, side outputs), but logs will list sources sorted.
- With one failing source the other sources still finish building (bounded, read-only probes); the first failure in name order is the one reported.
- A panicking source constructor now surfaces as an internal error from the blocking task instead of unwinding the run loop.
- `build_source_blocking` awaits its join handle immediately, so it detaches no task and needs no `ComponentScope` under the clippy policy #75 added.

## Testing

- `cargo test` (lib), on the merged tree: `streamling` 89 passed including two new tests for `build_source_providers` (every non-plugin source prepared; first failure in name order wins), `streamling-config` 31 passed, `streamling-connectors` 296 passed.
- `cargo clippy --all-targets --all-features -- -D warnings -A clippy::result_large_err` and `cargo fmt --check` clean.
- CI `e2e-tests` passed on the pre-merge head and runs again on the merge commit.
- Manual: the fake-endpoint runs above, plus plugin source/transform/sink pipelines, file sources, and the ClickHouse failure path on the built binary.

## Note on the red `claude-review` check

It fails on every PR in this repo since #115 switched the workflow to `pull_request_target` without `id-token: write`, and is unrelated to this diff. Details and a proposed one-line patch are in [this comment](https://github.com/goldsky-io/streamling/pull/126#issuecomment-5668154073).

https://claude.ai/code/session_01RnfQgpc2BczRSuTwyj7U6m

### DIFF
--- a/crates/streamling-config/src/app_config.rs
+++ b/crates/streamling-config/src/app_config.rs
@@ -7,6 +7,7 @@ use serde_derive::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::env;
 use std::fmt::Formatter;
+use std::sync::{Mutex, OnceLock};
 use std::time::Duration;
 
 fn default_sslmode() -> String {
@@ -303,8 +304,31 @@ impl std::fmt::Debug for KafkaConfig {
 impl KafkaConfig {
     /// Returns schema registry settings if a schema registry URL is configured.
     /// Returns None if no schema registry URL is set (e.g., when using JSON format).
+    ///
+    /// Settings are built once per (url, username, password) and cached for the
+    /// life of the process: building them constructs a `reqwest::Client`, which
+    /// parses the system CA bundle, and every Kafka source (plus each hybrid
+    /// source's Kafka phase) asks for one at startup. `SrSettings` is `Clone`
+    /// around an `Arc`'d client, so the clones share one connection pool too.
     pub fn get_schema_registry_settings(&self) -> Option<SrSettings> {
         let url = self.schema_registry_url.as_ref()?;
+        let key = (
+            url.clone(),
+            self.schema_registry_username.clone(),
+            self.schema_registry_password.clone(),
+        );
+
+        // The lock is held across the build on purpose: sources are constructed
+        // concurrently at startup, and a check-then-insert would let every one
+        // of them miss the cache at once and build its own client.
+        let mut cache = SCHEMA_REGISTRY_SETTINGS
+            .get_or_init(Default::default)
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
+        if let Some(settings) = cache.get(&key) {
+            return Some(settings.clone());
+        }
+
         let mut builder = SrSettings::new_builder(url.clone());
 
         if let (Some(username), Some(password)) = (
@@ -314,13 +338,19 @@ impl KafkaConfig {
             builder.set_basic_authorization(username, Some(password.as_str()));
         }
 
-        Some(
-            builder
-                .build()
-                .expect("failed to build schema registry settings from KafkaConfig"),
-        )
+        let settings = builder
+            .build()
+            .expect("failed to build schema registry settings from KafkaConfig");
+        cache.insert(key, settings.clone());
+        Some(settings)
     }
 }
+
+/// Process-wide cache behind [`KafkaConfig::get_schema_registry_settings`],
+/// keyed by (url, username, password).
+type SchemaRegistryKey = (String, Option<String>, Option<String>);
+static SCHEMA_REGISTRY_SETTINGS: OnceLock<Mutex<HashMap<SchemaRegistryKey, SrSettings>>> =
+    OnceLock::new();
 
 /// Compression codec applied by the Kafka sink's producer (librdkafka
 /// `compression.type`). Defaults to `lz4`, which is the historical built-in
--- a/crates/streamling-connectors/src/table_providers/clickhouse.rs
+++ b/crates/streamling-connectors/src/table_providers/clickhouse.rs
@@ -2087,6 +2087,35 @@ impl ClickHouseSourceExec {
     }
 }
 
+/// The one `reqwest::Client` behind every [`ClickHouseClient`] in the process.
+///
+/// Building a `reqwest::Client` is not cheap: the TLS backend parses the
+/// system CA bundle (twice, via openssl) on every build, roughly tens of
+/// milliseconds each. A pipeline used to build one per `ClickHouseClient` — a
+/// hybrid source alone constructs three (schema adapter, bounded source,
+/// offset provider) — which on a wide topology added up to over a second of
+/// pure CPU at startup and dominated `--validate` time. The pool settings below
+/// carry no per-connection state, and credentials, database and compression
+/// stay per `ClickHouseClient`, so one shared client (and connection pool) is
+/// equivalent and `Clone` on it is an `Arc` bump.
+static SHARED_HTTP_CLIENT: Lazy<reqwest::Client> = Lazy::new(|| {
+    // Configure HTTP client with optimized connection pooling settings
+    // These settings improve connection reuse and reduce latency for high-throughput workloads
+    reqwest::Client::builder()
+        // Keep idle connections in the pool for 90 seconds (default is also 90s, but explicit)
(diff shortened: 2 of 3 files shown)


rafadaguiar:
Merged `main` into this branch (`1b9d982`) — it was conflicted against #75 and #127. Resolution notes are in the merge commit message; the short version is that the registration loop still applies every builder #75 added (`with_scope`, `with_checkpoint_control`, `with_shutdown`), and only the constructors — the part that does I/O — run concurrently. The hybrid constructor's new `ComponentScope` parameter is created on the main thread and moved into the build task.

Re-verified locally on the merged tree: `cargo clippy --all-targets --all-features -- -D warnings -A clippy::result_large_err` and `cargo fmt --check` clean; unit tests `streamling` 89, `streamling-config` 31, `streamling-connectors` 296, all passing.

### Standing down on the `claude-review` check

`claude-review` is red here and is **not this PR's failure**. It fails before running any review logic:

```
error: Error message: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
Attempt 3 failed: Could not fetch an OIDC token. Did you remember to add `id-token: write` to your workflow permissions?
##[error]Action failed with error: Could not fetch an OIDC token.
```

Evidence that it is pre-existing and repo-wide:

- `431add1` (#115, Sep 8) switched `.github/workflows/ai-review.yml` to `pull_request_target`. That commit is this PR's *original base*.
- Every `AI Code Review` run since has failed — runs 480 through 487, across `claude/turbo-validate-perf-regression-nim8bn`, `fix/hybrid-projection-pushdown`, `feat/node-wait-starved`, `feat/node-wait-blocked`, `claude/kafka-clickhouse-source-issue-y7ccxs` and `maestro/…`, by four different authors. The last green runs (478, 479, Sep 7) predate the switch and ran on `pull_request`.
- #127 merged to `main` with this same check red.
- This PR touches no workflow files.

No fix exists on `main` yet, so there is nothing to port. The patch looks like one line — the job declares an explicit `permissions:` block, so everything unlisted is `none` and the runner never injects the OIDC env vars:

```diff
   claude-review:
     needs: [risk-scoring]
     if: github.event.pull_request.draft == false && github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     permissions:
       contents: read
       pull-requests: write
+      id-token: write
```

`streamling.yml` already does this for its AWS OIDC jobs. I have deliberately not made that change here — it is a CI fix unrelated to this diff and belongs in its own PR, which I'm happy to open if you want it.

### On the e2e question in the description

The description asked for `just e2e-test` before merging. That has since run in CI: `e2e-tests` passed on the pre-merge head. It will run again on the merge commit, and I'll report the result.

---
_Generated by [Claude Code](https://claude.ai/code)_

rafadaguiar:
CI is green on the merge commit (`1b9d982`) — reporting back on the e2e question from my last comment.

| Check | Result |
|---|---|
| `e2e-tests` | ✅ passed |
| `validate-and-test` | ✅ passed |
| `build-release-main` (x86_64, aarch64) | ✅ passed |
| `changes`, `risk-scoring` | ✅ passed |
| `claude-review` | ❌ repo-wide OIDC failure, see above |

So the e2e suite the description asked for has now run against the shutdown/drain work from #75 on this branch, and passes.

Re-ran the fake-endpoint measurement on the merged binary to confirm the speedup survived the merge (500 ms injected per call):

| Topology | before | after (merged) |
|---|---|---|
| 6 hybrid datasets → union → ClickHouse sink | 17.1 s | 2.8 s |
| 6 kafka datasets → union → ClickHouse sink | 4.0 s | 0.8 s |
| 1 hybrid dataset → ClickHouse sink | 3.2 s | 3.0 s |

All runs still return `is_valid: true` with identical warnings.

The merge-conflict state is cleared; the PR is mergeable and waiting on review from @kashive.

---
_Generated by [Claude Code](https://claude.ai/code)_

rafadaguiar:
👋 @kashive

---
_Generated by [Claude Code](https://claude.ai/code)_

rafadaguiar:
Two updates.

**Merged `main` (`9010303`).** File Source V2 (#136) replaced the file source's two constructors with a single `try_new` taking a `FileSourceReadMode`, which is the call this branch moved into `build_source_providers`, so the PR went conflicted. Ported, with two things worth a reviewer's eye:

- The read mode is built *inside* the build task, not before it. A bad `poll_interval` is a synchronous parse error; building the mode eagerly would let it short-circuit ahead of every concurrent build and quietly stop the name-order sort from deciding which error a failing topology reports. `reports_the_first_failure_in_name_order` pins that.
- `FileSourceReadMode::Bounded` carries the `CheckpointControl` that mints the source's terminal checkpoint, so `build_source_providers` takes that handle again — it had lost it when #75's hybrid wiring moved to the registration loop.

**Addressed the review's coverage gap (`4344a84`).** The AI review correctly noted that the existing tests only reach the file source, which builds inline on the async path, leaving the `spawn_blocking` closures and their error mapping untested. Added two tests, both using a ClickHouse source pointed at a port nothing listens on so a refused connection fails the constructor without waiting out a connect timeout (~1s total):

- `reports_a_blocking_pool_build_failure` — the constructor's own error survives the blocking task, rather than surfacing as the `JoinError` arm a panicking or cancelled task would produce.
- `orders_failures_across_both_build_paths` — name ordering decides across the two paths, not just within one, racing a blocking-pool source against an inline one.

Local gate on the pushed head: 91 `streamling` tests (up from 89), clippy with the CI flags, and `cargo fmt --check`, all clean.

Also noting that #137 removed the ai-review job, so the `claude-review` check I stood down on earlier no longer runs. `maestro / review` passed on the merge commit.

---
_Generated by [Claude Code](https://claude.ai/code)_

xiaom (approved):

PR #127, opened 2026-09-09
fix(connectors): honor projection pushdown in the hybrid source

bsidelinger912:
## Problem

Any pipeline that fast-scans a **column-sparse derived dataset** — one whose `transformSql` references only a subset of its root dataset's columns — fails physical planning. Today that set is exactly the four `<chain>.native_transfers` datasets (13 of raw_traces' 22 columns), which power wallet-feed native backfills (FOU-1173):

```
Internal error: Assertion failed: col.name() == matching_name: Input field name reward_type
does not match with the projection expression block_timestamp.
```

Minimal repro (fails at validation, no data needed; the identical pipeline with `start_at: latest` passes — `latest` resolves to a plain Kafka source, never a hybrid):

```yaml
sources:
  nt: { type: dataset, dataset_name: base_sepolia.native_transfers, version: 1.0.0, start_at: earliest, filter: "1=1" }
transforms:
  out: { type: sql, primary_key: id, sql: SELECT id, sender, recipient, amount FROM nt }
sinks:
  devnull: { type: blackhole, from: out }
```

## Root cause (two layers)

The projection DataFusion pushes into the hybrid scan is determined by the **dataset transformSql's referenced-column set** (transforms plan at their own boundary — the user's downstream SQL does not prune through it; the repro's assertion lists the transformSql's 13 columns, not the user query's 6).

1. **Plan time**: `HybridSourceExec` reported the provider's **full** schema — in `schema()` and in the `EquivalenceProperties` of its `PlanProperties` — even when a projection was pushed into `scan()`. DataFusion pairs projection expressions against schema fields **by index**: the 13 projected raw_traces columns put `block_timestamp` at projected index 12, while full-schema field 12 is `reward_type` — the two names in the assertion. This is independent of any column-order difference between storages (the first unit test reproduces it with identical bounded/unbounded schemas).
2. **Runtime**: the bounded `ClickHouseTableProvider::scan` ignores the projection argument entirely and emits every column in **ClickHouse DESCRIBE order**, which need not match the hybrid (Kafka) schema order the exec declares.

**Why this was latent:** every previously fast-scanned dataset's transformSql references **all** of its root's columns — e.g. `erc20_transfers` uses 10/10 raw_logs columns (`SELECT *` CTE + decode consuming `topics`/`data`) — so DataFusion never had a projection to push and the broken path was never exercised. It was *not* protected by column-order compatibility. `native_transfers` is the first dataset sparse over a wide root (raw_traces' fat `input`/`output` calldata columns are exactly what it exists to strip), so it trips the assertion 100% of the time.

The Kafka source already implements the correct pattern (`kafka.rs`): it projects its `PlanProperties` schema, decodes payloads directly into the projected schema, and when a source-level `filter` needs pruned columns it builds unprojected, filters, and re-applies the projection on top.

## Fix

Both contained in `HybridSourceExec`, mirroring the Kafka provider's contract:

- `new()` computes the **projected schema** (`provider.schema.project(indices)`) and uses it for `schema()` and `PlanProperties` (constructor is now fallible; both call sites propagate).
- The phase-forwarding loop **aligns every batch to the declared schema by column name** (`align_batch_to_schema`) before the existing ClickHouse u256 normalization — so neither inner provider's projection behavior nor its column ordering is load-bearing. Kafka-phase batches arrive already projected, so the fast path is a no-op clone; the alignment does real work only for the bounded ClickHouse phase. A missing column now produces a clear error instead of a DataFusion internal assertion.

Deliberately *not* done here: pushing the projection into the ClickHouse `SELECT` itself (perf follow-up — skips shipping `input`/`output` calldata for matched granules).

## Blast radius

Nothing running can regress:
- Every pipeline that enters the changed code path (hybrid = `start_at: earliest` + filter, per the dataset preprocessor) with a sparse projection failed at planning and therefore **never deployed** — there is no workaround population to break.
- Source-level filters on root-dataset columns (`topics`, `from_address`, …) are the normal mechanism and are untouched — filter planning is separate from projection on both phases.
- `start_at: latest` pipelines (e.g. the one existing native_transfers consumer) resolve to plain Kafka sources and never touch `HybridSourceExec`.
- Datasets whose transformSql references all root columns (everything except the four native_transfers datasets today) produce no projection, so their plans are byte-for-byte unchanged.

## Testing

- Two new unit tests, written first and failing on the exact pre-fix behaviors: `test_scan_with_projection_reports_projected_schema` (fails with *identical* schemas — proving order-matching can't fix this) and `test_execute_with_projection_aligns_reordered_phase_batches` (bounded mock ignores projection and emits reordered columns, mirroring the real ClickHouse provider).
- 36/36 hybrid tests, 291/291 `streamling-connectors` lib tests, `cargo fix`/`fmt`/`clippy -D warnings` clean; CI green including e2e.
- **Live verification** in the foundry kind cluster (engine image built from this branch, validator swapped via the agent image): a wallet-feed native-transfers backfill that previously failed planning fast-scanned `base_sepolia` raw_traces + raw_logs to completion, and a live native ETH transfer on base from a watched wallet flowed through the edge pipeline to the sink.

## Urgency

Wallet-feed **native transfers backfills are blocked on this landing fleet-wide** — apply-time validation runs the fleet engine, so no per-pipeline `imageTag` override can dodge it. (Known fallback if this stalls: re-author the native datasets to reference every raw_traces column, which suppresses the projection — at the cost of shipping calldata through every scan and into the public schema. Prefer the engine fix.)

Conflicts with #124 (unbounded-phase parallelism) in `HybridSourceExec::new` and the forwarding loop — noted there; landing this first keeps the rebase mechanical.

### DIFF
--- a/crates/streamling-connectors/src/table_providers/hybrid.rs
+++ b/crates/streamling-connectors/src/table_providers/hybrid.rs
@@ -881,7 +881,7 @@ impl TableProvider for HybridTableProvider {
             projection.cloned(),
             filters.to_vec(),
             limit,
-        )))
+        )?))
     }
 }
 
@@ -890,6 +890,12 @@ pub struct HybridSourceExec {
     inner: Arc<dyn ExecutionPlan>,
     provider: HybridTableProvider,
     cached_properties: Arc<PlanProperties>,
+    /// The schema this exec actually emits: the provider schema with the
+    /// pushed-down projection applied. Reporting the full provider schema
+    /// while a projection is in force makes DataFusion pair projection
+    /// expressions against full-schema fields by index and fail planning
+    /// ("Input field name X does not match with the projection expression Y").
+    schema: SchemaRef,
     projection: Option<Vec<usize>>,
     filters: Vec<Expr>,
     limit: Option<usize>,
@@ -902,9 +908,14 @@ impl HybridSourceExec {
         projection: Option<Vec<usize>>,
         filters: Vec<Expr>,
         limit: Option<usize>,
-    ) -> Self {
+    ) -> DataFusionResult<Self> {
+        let schema: SchemaRef = match projection.as_ref() {
+            Some(indices) => Arc::new(provider.schema.project(indices)?),
+            None => provider.schema.clone(),
+        };
+
         let cached_properties = PlanProperties::new(
-            EquivalenceProperties::new(provider.schema.clone()),
+            EquivalenceProperties::new(schema.clone()),
             inner.properties().partitioning.clone(),
             datafusion::physical_plan::execution_plan::EmissionType::Incremental,
             datafusion::physical_plan::execution_plan::Boundedness::Unbounded {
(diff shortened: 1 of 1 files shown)


bsidelinger912:
@claude review

rafadaguiar (approved):
Looks good to me.
Found some other issues while looking at this, but I'll address them in a followup pr.
This one is good to merge as it is.

PR #130, opened 2026-09-14
feat(connectors): Kafka producer statistics and a producer config escape hatch

rafadaguiar:
## Why

When a Kafka sink is slow there are two causes with opposite fixes, and today they are indistinguishable from the outside:

1. the pipeline is not feeding the producer fast enough, or
2. the broker/network is not draining it.

The only existing signal is librdkafka's `REQTMOUT` warning, which fires long after the pipeline has already stalled and says nothing about which case you are in. Diagnosing a slow sink currently means guessing, or rebuilding an image with ad-hoc logging.

## What

**1. Periodic producer statistics** (`ClientContext::stats`), off unless `statistics_interval_ms` is set, so this costs nothing by default.

One producer-wide line plus one line per broker:

- `queue_used_pct` / `queue_msgs_pct` — local producer queue utilisation. Near 100 means the sink is the bottleneck; near 0 means the producer is starved and the bottleneck is upstream. This alone separates the two cases above.
- `rtt_avg_ms` / `rtt_p99_ms` — broker round-trip time.
- `throttle_avg_ms` / `throttle_max_ms` — the broker's own `throttle_time_ms` from the ProduceResponse, i.e. a deliberate quota. Non-zero here is a definitive answer.
- `wire_kbps` / `reqs_per_s` / `avg_req_kb` — what is actually achieved on the socket, which distinguishes a bandwidth limit from a latency limit. `avg_req_kb x in_flight / rtt` is the achievable ceiling.
- `in_flight` / `outbuf_reqs` — pipelining depth, so a bandwidth-delay-product limit shows up as a low in-flight count against a full queue.
- `int_latency_avg_ms` / `outbuf_latency_avg_ms`, `req_timeouts_delta`, `retries_delta`, `connects_delta` / `disconnects_delta`.

**2. Per-partition backlog.** Every broker-level figure above is an aggregate, but Kafka preserves ordering per partition, so a single partition that stops draining throttles the whole sink while `queue_used_pct`, `rtt` and `wire_kbps` all still look reasonable. A per-tick summary makes that visible, plus a warning for each partition holding a backlog that transmitted nothing since the previous sample. Individual partitions are logged only when actually stalled, so a healthy sink stays quiet, and the list is capped so a broker-wide outage cannot turn one sample into hundreds of lines.

Two cases that would otherwise report false stalls are excluded: librdkafka's internal partition `-1`, which holds not-yet-assigned messages and never transmits, and the first sample after startup, which has no previous counter to diff against.

**3. `producer_overrides`** — a comma-separated `key=value` list applied after both the optimizer defaults and the per-sink settings, so broker-specific tuning (in-flight limits, request timeouts, batching) can be tried against a real deployment without cutting a new image. Malformed entries are logged and skipped rather than failing the sink, so a typo in a debugging knob cannot take a pipeline down.

## Notes

- Statistics are opt-in; with `statistics_interval_ms` unset the callback never fires.
- `connects`/`disconnects` are reported as deltas rather than cumulative counters so reconnect churn is visible in one line; the cumulative connect count is kept alongside for context.
- Caveat worth knowing when reading the output: if no responses arrive at all, librdkafka's `rtt` window may keep reporting its last value rather than degrading, so `rtt_avg_ms` should not be read in isolation — cross-check `wire_kbps` and `req_timeouts_delta`.

## Testing

Unit tests cover the override parser (including malformed entries), that statistics stay off unless configured, and the partition-stall rule (backlog with no transmission, the internal `-1` partition, and the no-previous-sample case). `cargo clippy --all-targets --all-features` is clean; the crate's 53 kafka tests pass.


kashive:
maestro comments seem legit and worth looking into.

btw you may want to explore my PR babysitting skill that automatically addresses the AI reviews https://gist.github.com/kashive/a038d5e0ddcf4d2c550b6a149c5d19ae

PR #121, opened 2026-09-04
feat(telemetry): emit node_wait{state="starved"} and dual-emit elapsed_compute

kashive:
## Summary
- Split out of #38 (2 of 3). Stacked on #120.
- Adds `streamling_node_wait_milliseconds_total{state="starved"}` for `WrappingExec` input-wait (`data.next().await`).
- Dual-emits the same remainder-carrying whole-ms span into `elapsed_compute` so existing dashboards stay valid and `busy = elapsed_compute - starved` does not go negative.
- EOF / upstream-error polls are not counted as starvation.
- `blocked` is **not** in this PR. The counter already carries `state` and `downstream_id=""` so #122 can add `state="blocked"` without a label-set change.

Part of https://linear.app/goldsky/issue/STRM-6325/streamling-per-nodeoperator-backpressure-attribution-metrics

**Stack:** #120 → **#121** → #122

After #120 merges, retarget this PR to `main`.

## Test plan
- [x] `cargo test -p streamling-core --lib starved`
- [x] `cargo test -p streamling-core --lib node_wait`
- [x] `cargo test -p streamling-core --lib busy_triad`
- [x] `just fix && just lint`
- [ ] e2e `test_source_starved_on_slow_upstream` (needs k3s; same test already passed locally on #38)

PR #133, opened 2026-09-14
feat(telemetry): per-node flow metrics for idleness and backpressure

rafadaguiar:
## Summary

Throughput graphs tell you a node stopped producing rows; they don't tell you why. This adds six per-node flow metrics that separate the three causes — waiting on the input, blocked on the output, or busily processing batches that happen to be empty.

| Metric | Type | Meaning |
| --- | --- | --- |
| `node_idle_wait` | Histogram (ms) | Waiting for the next input batch |
| `node_backpressure_wait` | Histogram (ms) | Blocked sending downstream |
| `node_backpressure_events` | Counter | One wait sample at or above the threshold |
| `node_empty_batch` | Counter | A batch with zero rows |
| `node_empty_streak` | Gauge | Consecutive empty batches; reset by a non-empty one |
| `node_inflight_buffered` | Gauge | Work held at this node boundary right now |

All six carry the node's standard tags plus `execution_kind` (`native` / `plugin`).

## Design

Emission goes through a new `telemetry::node_flow` helper module, so each operator adds a call instead of growing its own timing bookkeeping. Every helper takes `Option<&MetricsRecorder>` and is a no-op when telemetry was never initialized — instrumented paths stay silent in tests and tooling. `try_get_metrics_recorder` returns that `Option` without installing the warn-once no-op recorder that `get_metrics_recorder` does.

`node_backpressure_wait` gets its own bucket layout. The shared `DURATION_MS_BOUNDARIES` start at 100ms, which is far above a typical channel send, so it prepends fine-grained sub-100ms buckets while keeping the same high end — a wedged sink can block for minutes.

`node_inflight_buffered` is deliberately **local to one boundary**, not a global queue depth. Two nodes' values are not comparable as a queue; each says how much that node holds at that moment. `docs/node-flow-metrics.md` spells out the value at each site.

## Emission sites

`WrappingExec` and `WrappingDataSink` (idleness and emptiness only — the wrapper does not own the downstream send, so a backpressure sample there would attribute another node's stall to it), `spawn_marker_preserving_forwarder` (used by `CheckpointableExec`), `ExternalHandlerExec`, `WasmRunnerExec`, `PluginExec` (all six), and `HybridSourceExec`.

## Supporting changes

- `record_{count,gauge,time}_w_tag_slice` take `&[(&str, &str)]` so hot-path callers pass a stack array instead of allocating a `Vec` per emission. The existing `Vec`-taking methods delegate to them, so no call site had to change.
- `WasmRunnerNode` / `WasmRunnerExec` carry a `metric_metadata_id` (excluded from the node's identity impls, so adding telemetry cannot change planning), and `HybridTableProvider` gains a `with_metric_metadata_id` builder. Both now attribute to the composite registry key rather than a bare reference name.
- `process_plugin_metrics` takes an optional recorder and keeps draining its channel when telemetry is absent, rather than requiring one.
- Three channel sends that used `.unwrap()` now report the dropped receiver and end the stream instead of panicking.

## Testing

- New unit tests for the streak bookkeeping and for `send_batch_with_metrics` reporting delivery vs. a dropped receiver.
- `cargo test` green for `streamling-core` (576), `streamling-connectors` (296), `streamling` (87), plus the integration tests.
- `cargo clippy --workspace --all-targets` produces no new warnings.

## Docs

`docs/node-flow-metrics.md` (indexed from `docs/README.md`) covers the metric definitions, the threshold env var, emission sites, and how to read the in-flight gauge. `README.md` gets a short section in the telemetry chapter pointing at it.


kashive:
🤣 I have similar featured PR sitting in review for some time now 

https://github.com/goldsky-io/streamling/pull/121

